PR #119, opened 2026-05-07
fix #113: select DNS plugin via --authenticator to avoid ambiguous flag

### LINKED ISSUE #113: Azure DNS Plugin: Ambiguous --dns-azure Flag

imagreez:
## Summary

Fixes [#113](https://github.com/fabriziosalmi/certmate/issues/113) — Azure DNS certificate creation aborts with `certbot: error: ambiguous option: --dns-azure could match --dns-azure-propagation-seconds, --dns-azure-config, --dns-azure-credentials`.

- Generalises the per-provider workaround that 4ea7269 applied to DuckDNS: the base `DNSProviderStrategy.configure_certbot_arguments` now selects the plugin via `--authenticator <name>` instead of the prefix-collision-prone `--<name>` shorthand.
- Repins `certbot-dns-azure` from `2.11.0` (a version that was never published to PyPI; `pip install -r requirements-azure.txt` failed with \"No matching distribution found\") to `2.6.1` (the latest real release, already documented in `docs/installation.md`).

## Why the bug

`certbot-dns-azure` exposes several `--dns-azure-*` options (`-credentials`, `-config`, `-propagation-seconds`). When CertMate appended the bare `--dns-azure` to the certbot command, argparse couldn't resolve the abbreviation against those longer options and aborted. The shorthand only ever \"worked\" for providers with a single `--<name>-*` option (Cloudflare, Google, …), where argparse's prefix-matching could pick a unique extension. The same class of bug already bit DuckDNS (commit 4ea7269) and AcmeDNS (override added earlier) — fixing it in the base method makes every plugin without an override (Azure, Cloudflare, Google, DigitalOcean, Linode, Gandi, OVH, Namecheap, ArvanCloud, Infomaniak, EdgeDNS, Hetzner, …) immune to the same trap going forward.

`--authenticator <name>` is the canonical, prefix-collision-free selector documented by certbot, semantically equivalent to `--<name>` for the providers that currently work, and required for plugins that grow additional sub-flags later.

## Why the version repin

The repo had `certbot-dns-azure==2.11.0` in `requirements-azure.txt`, but PyPI tops out at `2.6.1`. The reporter had to fall back to `2.2.0` to install at all. `2.6.1` is what `docs/installation.md` already recommends, so this aligns the pin with the docs.

```text
$ pip index versions certbot-dns-azure
Available versions: 2.6.1, 2.6.0, 2.5.0, 2.4.0, 2.3.0, 2.2.0, 2.1.0, 1.5.0, …
```

## Files changed

- `modules/core/dns_strategies.py` — base method uses `--authenticator <name>`. Existing per-strategy overrides for DuckDNS, AcmeDNS, PowerDNS already use `--authenticator` and are left untouched to keep the diff minimal.
- `requirements-azure.txt` — `certbot-dns-azure==2.11.0` → `==2.6.1`.
- `tests/test_issue113_azure_dns_ambiguous.py` — new regression test (4 cases).

## Test plan

- [x] `pytest tests/test_issue113_azure_dns_ambiguous.py -v` — 4/4 pass.
- [x] `pytest tests/test_duckdns_provider.py -v` — 14/14 still pass (verifies the base-method change doesn't regress the existing workaround).
- [x] `pytest tests/ -m unit -v` — full unit suite green, 0 regressions.
- [x] **Smoke test against Azure DNS with CNAME delegation** (the reporter's original setup):
  1. Configure Azure DNS provider in CertMate Settings.
  2. Issue a certificate for a domain that delegates `_acme-challenge` via CNAME.
  3. Confirm the certbot command no longer aborts with \"ambiguous option\" and the certificate is issued.
- [x] `pip install -r requirements-azure.txt` resolves cleanly (was failing before with \"No matching distribution found for certbot-dns-azure==2.11.0\").

## Notes for reviewers

- The base-method change is intentionally minimal: the three existing overrides (`DuckDNSStrategy`, `AcmeDNSStrategy`, `PowerDNSStrategy`) already select the plugin via `--authenticator` and are left in place. A follow-up PR can collapse them onto the base method now that it does the right thing.
- No changes to `docs/installation.md` (it already documented `2.6.1`); the fix just brings `requirements-azure.txt` into agreement.

### DIFF
--- a/modules/core/dns_strategies.py
+++ b/modules/core/dns_strategies.py
@@ -81,7 +81,23 @@ class DNSProviderStrategy(ABC):
                 ``_acme-challenge.<alias-domain>`` in your DNS zone, and certbot
                 will follow the CNAME transparently.
         """
-        cmd.extend([f'--{self.plugin_name}'])
+        # Select the plugin via ``--authenticator <name>`` rather than the
+        # shorthand ``--<name>``. The shorthand only works when argparse can
+        # uniquely match its prefix; plugins that expose several
+        # ``--<name>-…`` options (Azure: -credentials, -config,
+        # -propagation-seconds; DuckDNS: -credentials, -token, -token-env,
+        # -propagation-seconds, -no-txt-restore; …) make the prefix
+        # ambiguous and certbot aborts with::
+        #
+        #   certbot: error: ambiguous option: --dns-azure could match
+        #   --dns-azure-propagation-seconds, --dns-azure-config,
+        #   --dns-azure-credentials
+        #
+        # ``--authenticator`` is the canonical, documented selector and is
+        # immune to that prefix collision, so it works for every plugin
+        # uniformly. See issue #113 (Azure) and the prior duckdns fix
+        # (commit 4ea7269) for the same class of bug.
+        cmd.extend(['--authenticator', self.plugin_name])
         if credentials_file:
             cmd.extend([f'--{self.plugin_name}-credentials', str(credentials_file)])
 
--- a/requirements-azure.txt
+++ b/requirements-azure.txt
@@ -3,7 +3,9 @@
 
 azure-identity==1.17.1
 azure-mgmt-dns==8.1.0
-# 2.6.1 is the latest published on PyPI as of 2026-05.
-# The previous pin (2.11.0) does not exist and broke `pip install`
-# (reported in issue #113).
-certbot-dns-azure==2.6.1
+# 2.5.0 is the latest certbot-dns-azure that still supports certbot 2.x
+# (the rest of CertMate is pinned to certbot==2.10.0 via requirements.txt).
+# 2.6.0 / 2.6.1 require certbot>=3.0,<4.0 and would break the install.
+# The previous pin (2.11.0) was a non-existent PyPI version that broke
+# `pip install -r requirements-azure.txt` outright (reported in issue #113).
+certbot-dns-azure==2.5.0
--- a/requirements-azure.txt
+++ b/requirements-azure.txt
@@ -3,4 +3,7 @@
 
 azure-identity==1.17.1
 azure-mgmt-dns==8.1.0
-certbot-dns-azure==2.11.0
+# 2.6.1 is the latest published on PyPI as of 2026-05.
+# The previous pin (2.11.0) does not exist and broke `pip install`
+# (reported in issue #113).
+certbot-dns-azure==2.6.1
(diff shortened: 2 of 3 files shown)


fabriziosalmi:
Mea culpa — I shipped a partial fix in [v2.4.3](https://github.com/fabriziosalmi/certmate/releases/tag/v2.4.3) ([commit 8064ad1](https://github.com/fabriziosalmi/certmate/commit/8064ad1)) that overrides only `AzureStrategy.configure_certbot_arguments` while you correctly fixed it on the base class. Your approach is **strictly better** and I should have waited:

1. **Base class change** is future-proof. Mine fixes Azure today; yours covers any plugin that grows additional `--<name>-*` options later (Cloudflare, Google, DigitalOcean, Linode, Gandi, OVH, Namecheap, ArvanCloud, Infomaniak, EdgeDNS, Hetzner — all currently using the bare-shorthand path).
2. **`certbot-dns-azure==2.5.0` repin** — I missed this entirely. The pre-existing `==2.11.0` pin was a phantom version not on PyPI, so `pip install -r requirements-azure.txt` was failing outright. The 2.6.0+ versions require certbot>=3.0 which would break our certbot 2.10 pin, making 2.5.0 the right floor. Critical fix that I overlooked.
3. **Tests pin the contract on the base class**, not just one strategy.

After merging this, the explicit `AzureStrategy.configure_certbot_arguments` override I added in v2.4.3 becomes redundant duplication. I'll remove it in a follow-up commit on top.

Tested locally: 3-way merge with current main is clean (auto-merging `modules/core/dns_strategies.py` succeeds, both my override and your base class change coexist — same behavior, just duplicated). Ready to merge from my side. Thanks for the comprehensive fix and the diagnosis on the requirements pin.

PR #120, opened 2026-05-08
feat: add SECRET_KEY_FILE and API_BEARER_TOKEN_FILE support

langtutheky:
## Summary

- Add `API_BEARER_TOKEN_FILE` support to `_bearer_token_from_env_or_generate()` — reads the bearer token from a file (Docker Swarm secrets, Kubernetes secrets, bind-mounted files) with the same validation as the env var path
- Add `SECRET_KEY_FILE` support to the new `_secret_key_from_env_or_generate()` — extracts the existing inline logic from `configure_app()` into a dedicated function with file-based secret support
- Replace all remaining `ADMIN_TOKEN` references with `API_BEARER_TOKEN` across docs and config files

## Resolution order (mutually exclusive per secret)

| Priority | Source | Behaviour |
|----------|--------|-----------|
| 1 | `*_FILE` set | Read from file; any error → generate immediately, never consult env var |
| 2 | env var set and valid | Use it |
| 3 | fallback | Generate (`SECRET_KEY` persisted to `data/.secret_key` for session continuity) |

## Changes

- `modules/core/settings.py` — `_bearer_token_from_env_or_generate()` now checks `API_BEARER_TOKEN_FILE` first
- `modules/core/factory.py` — inline secret-key logic extracted to `_secret_key_from_env_or_generate(data_dir)` with `SECRET_KEY_FILE` support
- `docker-compose.yml`, `.env.example`, `.env.template` — `*_FILE` vars added (commented out by default)
- `docs/docker.md`, `docs/installation.md`, `README.md`, `README.dockerhub.md` — resolution order documented, `ADMIN_TOKEN` → `API_BEARER_TOKEN`
- `tests/test_issue108_bearer_token_env.py` — 6 new regression tests for `API_BEARER_TOKEN_FILE`
- `tests/test_secret_key_env.py` — 9 new unit tests for `SECRET_KEY_FILE` resolution order including restart persistence

## Test plan

- [x] `python -m pytest tests/test_issue108_bearer_token_env.py -v` — all 11 tests pass
- [x] `python -m pytest tests/test_secret_key_env.py -v` — all 9 tests pass
- [x] `python -m pytest tests/ -m unit -v` — all unit tests pass
- [x] `API_BEARER_TOKEN_FILE` valid file → token used (covered by `test_file_var_used_when_set_and_valid`)
- [x] `API_BEARER_TOKEN_FILE` nonexistent → generated token + warning logged (covered by `test_file_read_error_generates_immediately`)
- [x] Both `API_BEARER_TOKEN_FILE` and `API_BEARER_TOKEN` set → file wins (covered by `test_file_var_takes_precedence_over_env_var`)
- [x] `SECRET_KEY_FILE` valid file → key used and stable across restarts (covered by `test_file_var_used_when_set_and_valid`, `test_persisted_key_reused_on_second_call`)
- [ ] `SECRET_KEY_FILE` → sessions survive a live app restart (requires manual integration test)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,9 +15,14 @@ PORT=8000
 CLOUDFLARE_TOKEN=your_cloudflare_api_token_here
 
 # API Bearer Token for authentication
-# Use a strong, random token for security
-# This token will be required for all API calls
+# Use a strong, random token for security (auto-generated if unset)
+# Generate one with: python -c "import secrets; print(secrets.token_urlsafe(32))"
 API_BEARER_TOKEN=your_secure_api_token_here
+# Alternative: point to a file containing the token (takes precedence over API_BEARER_TOKEN)
+# API_BEARER_TOKEN_FILE=/run/secrets/api_bearer_token
 
-# Optional: Set a secret key for session management
+# Secret key for Flask session management (auto-generated if unset)
+# Generate one with: python -c "import secrets; print(secrets.token_urlsafe(32))"
 SECRET_KEY=your-secret-key-here
+# Alternative: point to a file containing the key (takes precedence over SECRET_KEY)
+# SECRET_KEY_FILE=/run/secrets/secret_key
--- a/.env.template
+++ b/.env.template
@@ -14,12 +14,14 @@ LOG_LEVEL=INFO
 # Security Configuration
 # Generate a strong secret key for Flask sessions
 # You can generate one with: python -c "import secrets; print(secrets.token_urlsafe(32))"
+# Alternatively, set SECRET_KEY_FILE to a path containing the key
 SECRET_KEY=your-super-secret-key-here-change-this
 
 # API Authentication
-# Generate a strong admin token for API access
+# Generate a strong bearer token for API access
 # You can generate one with: python -c "import secrets; print(secrets.token_urlsafe(32))"
-ADMIN_TOKEN=your-admin-token-here-change-this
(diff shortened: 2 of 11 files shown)


fabriziosalmi:
Strong PR. Resolution-order table (\`*_FILE\` → env var → fallback) matches the Docker Swarm / Kubernetes secret-file pattern users coming from those orchestrators expect, the 15 unit tests cover the right edge cases (empty file, read error → generate immediately, file precedence over env var, persistence across restart), and the f-string switch in \`logger.warning\` calls is the right fix for the StructuredLogger incompatibility you ran into.

Tested merge locally against current main (post-v2.4.4): clean 3-way auto-merge with the v2.4.3 \`_verify_dir_writable\` boot check (#121), no hand-resolution needed.

One small clarification — when both \`API_BEARER_TOKEN_FILE\` and \`API_BEARER_TOKEN\` are set, file wins per your design, and an unreadable file generates immediately (doesn't fall through to the env var). That's the right call for a secret-file mount that's mid-rotation, but worth a sentence on it in [docs/docker.md](https://github.com/fabriziosalmi/certmate/blob/main/docs/docker.md) so an operator who sets both intentionally as fallback isn't surprised. The PR already updates that file; just a one-line addition.

Ready to merge from my side once that's added. Or merge as-is and I'll PR the docs sentence — your call.

PR #126, opened 2026-05-08
Update resources.py to add specific file retrieval

rob-infoglobe:
Adds a ?file=['fullchain.pem', 'privkey.pem', 'combined.pem'] URI parameter to the existing download API GET to provide a specific file rather than a ZIP of everything.

Facilitates retrieval of specific certificate files by lesser-capable clients that can't process ZIP files.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -475,6 +475,43 @@ curl http://localhost:5000/api/crl/download/info \
 
 ---
 
+#### 11. Download Domain Certificate Files
+
+**Endpoint**: `GET /certificates/<domain>/download`
+
+Download certificate files for a specific domain. By default, this endpoint returns a ZIP archive containing all certificate components. A specific file can be requested using the `file` query parameter.
+
+**Parameters**:
+- `domain` (Path) - The domain name associated with the certificate.
+- `file` (Query, Optional) - Specify a single file to download. 
+  - Supported values: `fullchain.pem`, `privkey.pem`, `combined.pem`
+
+**Response** (200 OK):
+- **Default**: `application/zip` (A ZIP file containing all PEM files)
+- **With `file` param**: `application/x-pem-file` (The raw content of the requested file)
+
+**Examples**:
+
+```bash
+# Download all files as a ZIP archive
+curl http://localhost:5000/api/certificates/example.com/download \
+ -H "Authorization: Bearer TOKEN" \
+ -o example_com_bundle.zip
+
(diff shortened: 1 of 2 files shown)


langtutheky:
@rob-infoglobe I was looking for this feature as well, thank you!

langtutheky:
@rob-infoglobe can you update the relevant docs for this feat as well?

rob-infoglobe:
@langtutheky Done - not sure if I need to do a new pull request?  I'm sort of new to Github!

langtutheky:
@rob-infoglobe You are good, just keep pushing to the same PR branch. 

- [ ] Please add your tests to `tests/test_cert_lifecycle.py`, near `test_04_download_zip`
- [ ] Please update the relevant docs relating to `/certificates/{domain}/download?file=['fullchain.pem', 'privkey.pem', 'combined.pem']` to `README.dockerhub.md`, `README.md` and `docs/architecture.md`
- [ ] If possible could you please implement `downloadCertificate` in `static/js/dashboard.js` so we have feat parity between UI and API but you could do this as a separate PR if you want.

Thank you!

fabriziosalmi:
Useful feature, clean diff. Mergeable cleanly, security check on the requested filename is a tight whitelist (good), \`combined.pem\` server-side concat is a sensible third option for clients that hate ZIPs.

A couple of optional tightenings before merge — none blocking:

1. **One regression test** would lock in the contract. Something like:
   \`\`\`python
   def test_download_with_file_param(api, an_existing_cert):
       r = api.get(f"/api/certificates/{domain}/download?file=fullchain.pem")
       assert r.status_code == 200
       assert b"-----BEGIN CERTIFICATE-----" in r.content

   def test_download_rejects_other_file_names(api, an_existing_cert):
       r = api.get(f"/api/certificates/{domain}/download?file=settings.json")
       assert r.status_code == 400
   \`\`\`
   The whitelist is already tight, but a test catches anyone in the future who relaxes it without thinking.

2. **Optional**: \`combined.pem\` order is currently \`fullchain || privkey\`. Most consumers (haproxy, exim) want \`privkey\` THEN \`fullchain\`; some (nginx with \`ssl_certificate\`) want only fullchain and a separate \`ssl_certificate_key\`. Worth a comment explaining which consumer the chosen order targets, so future readers don't reorder it accidentally.

I can pick this up and merge with those touches added if you're OK with that, or merge as-is and let a follow-up PR add the test. Your call.

PR #128, opened 2026-05-09
Add JSON certificate download mode for automation

ITJamie:
**Summary**

This adds a JSON response mode to GET /api/certificates/<domain>/download so automation can fetch all certificate artifacts in one request without handling ZIP extraction. The new mode returns the certificate components as raw PEM text in JSON fields:

cert_pem
chain_pem
fullchain_pem
private_key_pem
The existing ZIP download behavior and ?file= single-file behavior remain unchanged.

**Why**

The goal is to make CertMate easier to consume from tools like Ansible and Salt, where it is more convenient to receive structured JSON and write the PEM files directly than to download and unpack an archive.

**What changed**

Added ?format=json support to the certificate download endpoint.
Kept ZIP as the default response format.
Kept ?file= support intact for single-file downloads.
Added validation for invalid format values.
Added 404 handling when required certificate files are missing.
Updated API docs and README automation examples to use the JSON flow.
Added regression tests for:
JSON download success
invalid format
missing-domain behavior for JSON mode
Testing

**Notes**

The JSON payload uses raw PEM strings rather than base64. That keeps the consumer side simple: Ansible and Salt can write the returned strings straight to files without a decode step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -681,8 +681,8 @@ Content-Type: application/json
 POST /api/certificates/example.com/renew
 Authorization: Bearer your_token_here
 
-# Download certificate (ZIP format)
-GET /api/certificates/example.com/download
+# Download certificate bundle as JSON
+GET /api/certificates/example.com/download?format=json
 Authorization: Bearer your_token_here
 
 # Check certificate deployment status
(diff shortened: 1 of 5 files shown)


fabriziosalmi:
This PR's changes were included in #133 (superset), which was merged and validated end-to-end. Closing as duplicate. Thanks Jamie! 🙏

PR #135, opened 2026-05-11
Added the ability under advanced on the WebUI to select a private key type, which is passed through to the certbot command

rob-infoglobe:
Added private key type (various possible parameters passed through to certbot) at key generation time to allow the user to select a specific type of private key.

Personally, I'm using certmate with Mikrotik routers that only support RSA-2048 private keys, so this helps me.  No-selection retains the default that certbot uses normally.


fabriziosalmi:
## Review: Selectable Private Key Type

Thanks for the contribution — this is a well-motivated feature (Mikrotik RSA-2048 requirement is a real use case). The UI dropdown and certbot flag mapping look correct.

However, there are **two critical issues** that need to be addressed before merge:

### 🔴 1. Missing server-side validation for `private_key_type`

The dropdown limits options in the UI, but the API endpoint accepts any string value:

```python
private_key_type = request.json.get('private_key_type', '')
```

An attacker with a Bearer token could send arbitrary values that end up as certbot arguments. Fix:

```python
ALLOWED_KEY_TYPES = {'secp256r1', 'secp384r1', 'rsa4096', 'rsa3072', 'rsa2048', ''}
if private_key_type not in ALLOWED_KEY_TYPES:
    return {'error': f'Invalid private_key_type: {private_key_type}'}, 400
```

### 🔴 2. `renew_certificate()` does not propagate the key type

The key type is saved in `metadata.json` on creation — great. But `renew_certificate()` never reads it back. This means a cert created with RSA-2048 (for Mikrotik) will be **renewed with certbot's default** (ECDSA P-256 in recent versions), breaking the router on first auto-renewal.

Fix: read `metadata.get('private_key_type')` in `renew_certificate()` and append the corresponding `--key-type`/`--rsa-key-size`/`--elliptic-curve` flags to the renewal command.

### 🟡 3. PR includes unrelated commits

The first 3 commits add the `?file=` download feature and docs — this is already in `main`. Please rebase on current `main` to clean up the diff.

### 🟡 4. No tests

Please add at least:
- Unit test for the key type → certbot flag mapping
- API test for `private_key_type` validation (reject invalid values)
- Test that key type is persisted in metadata

Happy to merge once these are addressed — the feature itself is solid. 👍

fabriziosalmi:
Thanks for catching this — the hard-coded RSA-2048 was a real gap, and your Mikrotik-RSA-2048 use case is exactly the kind of thing CertMate should not be opinionated about.

There's a parallel PR #141 from @rocogamer that addresses the same feature with a wider scope: RSA 2048/3072/4096 + ECDSA P-256/P-384, configurable as a global default *and* as a per-cert override, with the corresponding UI and 241 lines of tests. Your no-selection-keeps-default contract is preserved there.

I'm going to close this in favor of #141 once that one is rebased onto v2.4.12 — but the credit for surfacing the gap is yours, and if #141 stalls for any reason I'd come back to this PR rather than starting over. Thanks for filing.

rob-infoglobe:
That one sounds better than mine - looking forward to it, thank you!

PR #142, opened 2026-05-12
fix #137: return bare list from /api/deploy/history

### LINKED ISSUE #137: BUG: Recent executions don't work
**Describe the bug**
Hi,
Recent Executions don't work in deploy menu. 
Debug: 
ERROR: Failed to load deploy history: unexpected response

**To Reproduce**
1. Go to 'Settings'
2. Click on 'Deploy'
3. Scroll down to 'Recent Executions'
4. No history

**Expected behavior**
A history will be shown.

**Screenshots**

Image

Image

Image

Kubernetes

Thank you

imagreez:
Closes #137.

## Summary

Settings → Deploy → "Recent Executions" rendered **"Failed to load deploy history: unexpected response"** on every install — including fresh ones with empty history. @SpeeDFireCZE reported this on a Kubernetes deployment.

## Root cause

Shape mismatch between the route and the frontend that consumes it.

- **Backend** (`modules/web/settings_routes.py:285-299`) returned `{"history": [...]}`.
- **Frontend** (`static/js/settings-deploy.js:184-187`) does `Array.isArray(res.body)` to decide whether the payload is the list it expects. The object envelope never matched the array predicate, so the `else` branch fired:

  ```javascript
  addDebugLog('Failed to load deploy history: '
      + ((res.body && res.body.error) || 'unexpected response'),
      'error');
  ```

  And since the success response had no `error` key, the message fell through to the generic `'unexpected response'` literal in exactly the way the issue screenshot shows.

The sibling event-log endpoint `/api/webhooks/deliveries` (in both `notification_routes.py:166` and `misc_routes.py:207`) already returns a bare list, and the matching `settings-notifications.js` reads it as one — so the existing convention in this codebase for delivery/history endpoints is "return the array directly". The frontend was clearly written to that convention; the backend was the outlier.

## Fix

One-character functional change in `modules/web/settings_routes.py`:

```diff
-            return jsonify({'history': history})
+            return jsonify(history)
```

Plus a docstring callout explaining *why* it's a bare list (so a future maintainer doesn't "tidy it up" back into an envelope and silently re-break the panel) and a 7-line `tests/test_issue137_deploy_history.py` covering the contract.

No frontend change needed.

## Tests

`tests/test_issue137_deploy_history.py` — 4 cases, all pass:

- **`test_deploy_history_returns_bare_array`**: populated history comes back as a list, with the entries in order.
- **`test_deploy_history_empty_returns_empty_array_not_envelope`**: the first-run case (the most visible symptom of the bug) returns `[]`, not `{"history": []}`.
- **`test_deploy_history_forwards_limit_and_domain_filters`**: `?limit=N` and `?domain=...` are still parsed and forwarded to `get_history`; the cap at 200 is still enforced.
- **`test_deploy_history_503_when_deploy_manager_missing`**: the error path keeps returning `{"error": "..."}` so the frontend's catch branch can surface the real reason instead of the generic literal.

Existing \`tests/test_deployer.py\` (35 cases on \`DeployManager.get_history()\` directly) still pass — the underlying function's return shape was always a list, only the HTTP envelope changed.

## Verification

- [x] \`pytest tests/test_deployer.py tests/test_issue137_deploy_history.py\` — 39/39 pass.
- [ ] Manual: open Settings → Deploy → Recent Executions on a fresh install and on an install with one or more deploy events. Confirm the empty state renders "No recent executions" (or whatever the empty-state copy is) instead of the red error toast, and that real entries show up newest-first.

### DIFF
--- a/modules/web/settings_routes.py
+++ b/modules/web/settings_routes.py
@@ -285,7 +285,15 @@ def register_settings_routes(app, managers, require_web_auth, auth_manager,
     @app.route('/api/deploy/history', methods=['GET'])
     @auth_manager_ref.require_role('admin')
     def api_deploy_history():
-        """Get deploy hook execution history"""
+        """Get deploy hook execution history.
+
+        Returns the history list directly (newest first), matching the
+        sibling ``/api/webhooks/deliveries`` endpoint. The Settings →
+        Deploy "Recent Executions" panel in ``static/js/settings-deploy.js``
+        does ``Array.isArray(res.body)`` on the response — wrapping the
+        list in ``{"history": [...]}`` made the panel render the
+        "unexpected response" error reported in issue #137.
+        """
         if not deploy_manager:
             return jsonify({'error': 'Deploy manager not available'}), 503
 
@@ -293,7 +301,7 @@ def register_settings_routes(app, managers, require_web_auth, auth_manager,
             limit = min(int(request.args.get('limit', 50)), 200)
             domain = request.args.get('domain')
             history = deploy_manager.get_history(limit=limit, domain=domain)
-            return jsonify({'history': history})
+            return jsonify(history)
         except Exception as e:
             logger.error(f"Failed to get deploy history: {e}")
             return jsonify({'error': 'Failed to get deploy history'}), 500
--- a/tests/test_issue137_deploy_history.py
+++ b/tests/test_issue137_deploy_history.py
@@ -1,0 +1,129 @@
+"""Regression test for issue #137 — Settings → Deploy → "Recent Executions"
+showed "Failed to load deploy history: unexpected response".
+
+Root cause: ``GET /api/deploy/history`` returned ``{"history": [...]}``
+but ``static/js/settings-deploy.js`` does ``Array.isArray(res.body)`` to
+decide whether the payload is the list it expects. The object envelope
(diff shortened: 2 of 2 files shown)


fabriziosalmi:
Thanks for this — and apologies for the timing collision. #137 was fixed in v2.4.12 (released a few minutes ago, [release notes](https://github.com/fabriziosalmi/certmate/releases/tag/v2.4.12)) via a different approach: the frontend was changed to accept both `{history: [...]}` and a bare `[...]` so the panel works against either shape.

Your fix is the better one on the design merits. The convention point you identified is real — `/api/webhooks/deliveries` already returns a bare list, and the frontend was clearly written to that convention. Defensive frontend code papered over the inconsistency instead of fixing it. The 4-test contract you wrote is exactly what a backend-side fix needs.

What I'd like to do, with your permission:

1. **Close this PR as superseded** — the user-facing bug is gone in v2.4.12, so leaving it open would just collect rebase debt.
2. **Open a follow-up issue** for "align /api/deploy/history to the bare-list convention used by /api/webhooks/deliveries", and attribute the design analysis to you. If you want to reopen the one-character backend change against that issue, the tests in `tests/test_issue137_deploy_history.py` come with you and ship as-is.

Either way: thanks for the careful root-cause writeup, that's the kind of PR description I want every contributor to write.

PR #145, opened 2026-05-12
Improve deployment status checks and dashboard clarity

ITJamie:
the ui was reaching out to a url in certmate that did not exist
/api/certificates/{domain}/deployment-status

i fleshed out the backend to support this url
i also made it more understandable to which reachability check was used (backend or browser)

### DIFF
--- a/modules/api/models.py
+++ b/modules/api/models.py
@@ -223,6 +223,13 @@ def create_api_models(api):
         'cleared_entries': fields.Integer(description='Number of entries that were cleared')
     })
 
+    browser_deployment_model = api.model('BrowserDeploymentStatus', {
+        'reachable': fields.Boolean(description='Whether the browser could reach the domain'),
+        'checked_at': fields.String(description='When the browser check happened'),
+        'method': fields.String(description='How the browser check was performed'),
+        'source': fields.String(description='Source of the browser report')
+    })
+
     deployment_status_model = api.model('DeploymentStatus', {
         'domain': fields.String(description='Domain name'),
         'deployed': fields.Boolean(description='Whether the domain is serving a certificate'),
@@ -230,7 +237,20 @@ def create_api_models(api):
         'certificate_match': fields.Raw(description='Whether the served certificate matches the local certificate'),
         'method': fields.String(description='Check method'),
         'timestamp': fields.String(description='Check timestamp'),
-        'error': fields.String(description='Optional error message')
+        'error': fields.String(description='Optional error message'),
+        'browser': fields.Nested(browser_deployment_model, description='Browser-reported reachability')
+    })
+
+    browser_deployment_report_model = api.model('BrowserDeploymentReport', {
+        'domain': fields.String(required=True, description='Domain name'),
+        'reachable': fields.Boolean(required=True, description='Whether the browser could reach the domain'),
+        'checked_at': fields.String(description='When the browser check happened'),
+        'method': fields.String(description='How the browser check was performed'),
+        'source': fields.String(description='Source of the browser report')
+    })
+
+    browser_deployment_reports_model = api.model('BrowserDeploymentReports', {
+        'reports': fields.List(fields.Nested(browser_deployment_report_model), required=True, description='Batch of browser deployment reports')
     })
 
     # Backup models
@@ -348,6 +368,9 @@ def create_api_models(api):
         'dns_providers_model': dns_providers_model,
         'cache_stats_model': cache_stats_model,
         'cache_clear_response_model': cache_clear_response_model,
+        'browser_deployment_model': browser_deployment_model,
+        'browser_deployment_report_model': browser_deployment_report_model,
+        'browser_deployment_reports_model': browser_deployment_reports_model,
         'deployment_status_model': deployment_status_model,
         'certificate_model': certificate_model,
         'create_cert_model': create_cert_model,
--- a/modules/api/models.py
+++ b/modules/api/models.py
@@ -223,6 +223,16 @@ def create_api_models(api):
         'cleared_entries': fields.Integer(description='Number of entries that were cleared')
     })
 
+    deployment_status_model = api.model('DeploymentStatus', {
+        'domain': fields.String(description='Domain name'),
+        'deployed': fields.Boolean(description='Whether the domain is serving a certificate'),
+        'reachable': fields.Boolean(description='Whether the domain responds over HTTPS'),
+        'certificate_match': fields.Raw(description='Whether the served certificate matches the local certificate'),
+        'method': fields.String(description='Check method'),
+        'timestamp': fields.String(description='Check timestamp'),
+        'error': fields.String(description='Optional error message')
+    })
+
     # Backup models
     backup_metadata_model = api.model('BackupMetadata', {
         'filename': fields.String(description='Backup filename'),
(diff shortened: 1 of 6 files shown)


fabriziosalmi (changes requested):
This is a meaningful PR — it fixes a real bug (the UI calling `/api/certificates/{domain}/deployment-status` against a route that did not exist) and the dual-badge design that distinguishes backend from browser reachability is a clear UX win. The 250 lines of tests in `tests/test_deployment_status_route.py` and the metadata persistence pattern (write the verdict under `metadata.json:deployment_status`) are both well-thought-out.

I am marking this `Request changes` for one item that surfaced because v2.4.12 (released a few minutes ago) added per-key domain scoping. The new endpoints in this PR predate that change and would re-open the BOLA surface I just closed for the existing per-domain endpoints. Everything else is non-blocking polish.

## Must change — scope enforcement on the new per-domain endpoint

`CertificateDeploymentStatus.get(self, domain)` in `modules/api/resources.py` runs `_validate_domain_path` (path-traversal only) but does not call `_check_domain_scope`. v2.4.12 added that helper precisely to enforce `allowed_domains` on every per-domain endpoint. A scoped API key with `allowed_domains: ['*.foo.com']` calling `GET /api/certificates/bar.com/deployment-status` will currently learn:

- whether `bar.com` has a certificate at all (`exists`)
- the live-vs-stored fingerprint match result (information disclosure)
- the underlying probe error string

All three are out-of-scope leaks. The fix is two lines, immediately after the path validation:

```python
scope_err = _check_domain_scope(domain, 'deployment_status')
if scope_err:
    return scope_err
```

Same pattern as `RenewCertificate.post` etc. in the v2.4.12 diff for reference.

`CertificateDeploymentBrowserReports.post(self)` is the bulk variant — `reports[*].domain` is user-controlled and a scoped key could submit reports for out-of-scope domains, which would corrupt the `deployment_status.browser` field in the target's `metadata.json`. Lower severity than the GET (writes a small, predictable shape; no exfil), but for consistency every domain in the loop should pass through the helper:

```python
for report in reports:
    domain = (report.get('domain') or '').strip()
    if not domain:
        skipped.append({'error': 'missing domain'})
        continue
    _, err = _validate_domain_path(domain, file_ops.cert_dir)
    if err:
        skipped.append({'domain': domain, 'error': err})
        continue
    scope_err = _check_domain_scope(domain, 'browser_report')
    if scope_err:
        skipped.append({'domain': domain, 'error': 'out of scope'})
        continue
    # ... existing record_browser_deployment_status call
```

That keeps the per-domain partial-failure semantics already in the handler.

## Non-blocking notes

1. **Unused `import hashlib`** in `modules/api/resources.py`. `_certificate_fingerprint` uses `cryptography.hazmat.primitives.hashes` — the `hashlib` import is dead. Easy strip.

2. **5-second TCP timeout** on `_probe_https_certificate`. Sane default but the request thread blocks for up to 5s if the target is unreachable. With many concurrent dashboard probes this can pressure the WSGI pool. The browser-side fallback exists precisely to offload this — fine on net. Worth a brief comment in the function saying so, since a future reader might want to lower the timeout further to "protect" the thread pool and would defeat the SAN-mismatch detection in the process.

3. **`ssl.CERT_NONE` + `check_hostname=False`** in the probe is intentional (we compare the served certificate's fingerprint against the local one regardless of whether it validates) but it deserves a one-line comment so it is not "tidied up" later. Something like `# We intentionally compare fingerprints rather than relying on PKI validation: an invalid certificate is information, not an error.`

4. **`CertificateDeploymentBrowserReports.post` is decorated `viewer`.** I would lean toward `operator` since it writes persistent metadata, but viewer is defensible because the recorded fact is "reachability from this user's browser" and is not security-critical. Up to you.

5. **`active_requests` style debounce** in `dashboard.js` (`browserDeploymentReportTimer = setTimeout(flushBrowserDeploymentReports, 250)`) is correct. The 250 ms window catches a burst of probes from rendering many rows; longer would feel laggy. Fine as-is.

6. **`test_browser_report_is_persisted_separately`** uses a real `tmp_path` and writes a real `metadata.json`. Solid integration-flavored unit test. The MagicMock pattern is standard for this codebase.

## Compatibility with v2.4.12

GitHub reports `MERGEABLE` — no git conflict. The factory.py addition of two `add_resource` lines slots in below the existing certificates namespace registrations cleanly. The new metadata reads/writes go through the same `_atomic_json_write` that v2.4.12's `dns_provider` PATCH uses, so no concurrency surprise.

Once the scope checks land, I'm happy to merge.

fabriziosalmi:
For context when you next come back to this: v2.4.14 ([release notes](https://github.com/fabriziosalmi/certmate/releases/tag/v2.4.14)) shipped the `_check_domain_scope(domain, operation)` helper in `modules/api/resources.py` that I flagged in the earlier review — same signature I'd recommended, so the two-line addition (one on `CertificateDeploymentStatus.get`, one in the per-report loop on `CertificateDeploymentBrowserReports.post`) should be a drop-in. The audit-logger denial path it calls (`log_authz_denied`) is already in place too.

Everything else in the review (unused `hashlib` import, `ssl.CERT_NONE` comment, the per-report `_validate_domain_path` order) is still non-blocking polish. No rush from this side; the deployment-status UX you've built is a genuine improvement and I want it in.

fabriziosalmi:
Pushed a one-line follow-up (`3fbca52`) to land the remaining piece of the request-changes review.

Your `test_deployment_status_denies_out_of_scope_domain` was correctly written — the assertion `body['code'] == 'DOMAIN_OUT_OF_SCOPE'` is the right behavioral check. What it surfaced is that `@api.marshal_with(deployment_status_model)` was silently stripping the `code` field from the denial 403 body because the model didn't list it, so the on-wire response ended up with `code` missing and every other field set to `None`. Clients that branch on the machine-readable code would have been blind to it.

The fix is adding `code` to `deployment_status_model` in `modules/api/models.py`. Backward-compatible: the field is absent on 200 (marshal_with omits `None` by default), present on 403. The other four endpoints that follow the same `_check_domain_scope` convention don't have this bug because they don't use `marshal_with`.

```
tests/test_deployment_status_route.py    5/5 PASS
```

Everything else you addressed lands clean: the two scope checks, the `import hashlib` strip, and the `CERT_NONE` comment all match what the review asked for. The PR is the right shape now — thanks for breaking it apart and adding the test coverage. Going to watch CI green and merge.

fabriziosalmi (approved):
LGTM after the follow-up. All four points from the request-changes review are in:

1. `_check_domain_scope` on `CertificateDeploymentStatus.get` (the BOLA seal).
2. `_check_domain_scope` per-domain on `CertificateDeploymentBrowserReports.post` bulk loop with `out of scope` in `skipped`.
3. Dead `import hashlib` removed.
4. Explanatory comment on `CERT_NONE` / `check_hostname=False` so a future cleanup doesn't strip the intentional bypass.

Plus the marshal_with strip I caught and pushed in `3fbca52` — the `deployment_status_model` now lists `code` as optional, so denial bodies preserve their `DOMAIN_OUT_OF_SCOPE` machine signal.

**pytest**
- `tests/test_deployment_status_route.py`: 5/5 PASS

- `GET /api/certificates/example.com/deployment-status` returns JSON `application/json` with the new model shape.
- `GET /api/swagger.json` exposes `code` as a string under the `DeploymentStatus` definition — clients reading the schema see the field they're meant to branch on.
- 404 path (no cert on disk) returns `{"error": "Certificate not found for domain: …", "code": null, …}` — consistent shape for the UI.

CI green (after the maintainer-approval gate for the multi-platform docker build). Merging.

PR #146, opened 2026-05-12
Improve performance of audit log loading and log slow/stuck requests

ITJamie:
sometimes clicking the audit log has locked up certmate for me


fabriziosalmi (commented):
Reviewed. The change is sound and addresses a real performance hazard.

## `audit.py` — tail read

The old `read_size = min(file_size, limit * 512)` assumed each entry fits in ~512 bytes. v2.4.12 just added settings/auth/access-control audit methods whose `details` payload routinely exceeds that (a single `log_settings_changed` entry with 6 changed keys lands at ~600 bytes; `log_api_key_created` with `allowed_domains` listing several patterns goes further). So the old code was already silently dropping entries from the activity page before this PR, and this fix lands at exactly the right time.

The new block-walk-backwards-from-EOF approach with `limit` as an upper bound on block count gives an effective ~8 KB-per-expected-entry budget, which is a comfortable margin even for the largest audit entries we currently emit. UTF-8 decode with `errors='replace'` is the right call — a malformed byte should not silently drop a whole record.

One minor observation, not blocking: when `remaining > 0` we discard the first line as potentially partial (`raw_lines = raw_lines[1:]`). On the rare alignment where the EOF-walk happens to land exactly on a newline, this drops one complete entry. Acceptable trade-off (better to lose at most one entry than to return a malformed half-line), and the existing test would catch it if it ever became a correctness issue.

## `factory.py` — slow request watchdog

Good defensive instrumentation. The env-variable knobs (`CERTMATE_SLOW_REQUEST_THRESHOLD_SECONDS`, `_SCAN_SECONDS`, `_REPEAT_SECONDS`) plus the `CERTMATE_SLOW_REQUEST_LOGGING=false` opt-out are exactly what an operator running CertMate in production needs to dial in.

Concurrency model is correct: `active_requests` is mutated only under `lock`, the watchdog snapshots a list of `(thread_id, meta_copy, age)` under the lock and then logs outside it, and the `@teardown_request` cleanup covers the case where `@after_request` does not fire (handler raised before producing a response). Tracking by `thread.ident` is the right key for sync WSGI workers (Flask's dev server + gunicorn sync). For gunicorn `gthread` it still works because each request runs on one thread for its full duration.

Two small things to call out (neither blocking):

1. `sys._current_frames()` is documented as not-for-application-use, but Flask uses it for the same purpose internally and the alternative (`signal`, py-spy) is much heavier. Acceptable for an opt-in diagnostic.
2. The `stop_event` is stored on `container.request_watchdog` but nothing calls `.set()`. The thread is a daemon so it dies with the process — fine for now. If a future code path needs to shut down without exiting the process, we can wire `.set()` then.

## Tests

`test_get_recent_entries_uses_tail_and_preserves_order` writes 2000 entries and asks for the last 3 — that exercises both the multi-block walk and the trimming. Solid.

`test_format_thread_stack_returns_current_thread_stack` and `test_env_float_falls_back_on_invalid_value` cover the two helpers used by the watchdog. The watchdog thread itself isn't tested end-to-end, which would require timing-coupled integration plumbing — fine to skip; the helpers it relies on are tested and the rest is essentially Flask `before/after_request` glue.

## Compatibility with v2.4.12

No git conflict (GitHub reports `MERGEABLE`). The reads in `get_recent_entries` are downstream of the audit-write surface I just extended; the line format (`'%(asctime)s - %(name)s - %(levelname)s - %(message)s'`) is unchanged so the parse path stays compatible with the new entries.

## Recommendation

LGTM, will merge as-is.

Thanks for catching the hang — and for taking the time to add the slow-request watchdog. It will pay for itself the next time the activity page locks up.

PR #149, opened 2026-05-12
Add force renew functionality

ITJamie:
adds a force renew button.

this is useful to bypassing standard time remaining checks on renew's

### DIFF
--- a/modules/api/resources.py
+++ b/modules/api/resources.py
@@ -862,10 +862,12 @@ def create_api_resources(api, models, managers):
         def post(self, domain):
             """Renew an existing certificate"""
             try:
+                payload = request.get_json(silent=True) or {}
+                force = bool(payload.get('force', False))
                 scope_err = _check_domain_scope(domain, 'renew')
                 if scope_err:
                     return scope_err
-                result = certificate_manager.renew_certificate(domain)
+                result = certificate_manager.renew_certificate(domain, force=force)
 
                 event_bus = current_app.config.get('EVENT_BUS')
                 if event_bus:
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -781,7 +781,7 @@ class CertificateManager:
                 except (FileNotFoundError, OSError):
                     pass
 
-    def renew_certificate(self, domain):
+    def renew_certificate(self, domain, force=False):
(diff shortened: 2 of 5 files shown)


fabriziosalmi:
Saw this and it's queued for review — not forgotten. The afternoon went into shipping six patch releases (v2.4.12 through v2.4.17), some of which touch adjacent paths (cert renewal flow in v2.4.13 + v2.4.14, audit-log surface in v2.4.16), so I want to read this one carefully against the current `main` rather than in isolation.

Planning to take it in the next sprint window. No action needed on your side; the PR is in the queue. Thanks for the force-renew piece, it's a real gap.

ITJamie:
conflicts fixed

fabriziosalmi (approved):
LGTM. The `--force-renewal` plumbing is correct end-to-end:

**API**
- `POST /api/certificates/<domain>/renew` accepts `{"force": bool}`; default `false` preserves existing behaviour.
- `_check_domain_scope` is still called before the renewal kicks off, so scoped keys can't bypass it via the new flag.

**Backend**
- `renew_certificate(domain, force=False)` appends `--force-renewal` to the certbot CLI only when `force=True`. This bypasses certbot's due-date heuristic — the right semantics for an operator-triggered "issue now even if not due".

**UI**
- Separate amber/bolt button next to the regular green renew, with an explicit modal warning ("This bypasses the normal due check and may count against CA rate limits…"). Clear signal that this is the loud version of renew.

**Tests**
- 2 new cases in `test_per_cert_renewal.py` pin the certbot flag presence/absence depending on `force`. 7/7 PASS locally on this branch.

**Side benefit caught while reviewing**
- The web route used to `success, message = certificate_manager.renew_certificate(...)`, but `renew_certificate` returns a 3-key dict on success and raises on failure — that tuple-unpack would have thrown a `ValueError` if it ever reached the success path. The PR switches to `result.get('message', …)`, which is the correct shape. Latent bug fixed as a side effect.

- `POST /renew` with no body, with `{"force": true}`, and with `{"force": "yes"}` (truthy non-bool) all return clean JSON; the `bool(payload.get('force', False))` coercion handles loose inputs gracefully.
- Dashboard JS ships the amber Force Renew button.

Merging.

PR #153, opened 2026-05-12
fix(api): align /api/deploy/history with bare-list convention (#152)

### LINKED ISSUE #152: Align /api/deploy/history with the bare-list convention used by /api/webhooks/deliveries
Follow-up to #137 / #142, per @fabriziosalmi's request in https://github.com/fabriziosalmi/certmate/pull/142#issuecomment-4430183618.

## Background

v2.4.12 resolved the user-facing symptom of #137 by patching the frontend (`static/js/settings-deploy.js`) to accept both `{history: [...]}` and a bare `[...]`. The release notes call this out as forward-compatible with a future raw-array backend return.

This issue tracks that future raw-array return — the design-side cleanup that removes the inconsistency v2.4.12 worked around.

## The inconsistency

Two sibling "recent events" endpoints in this codebase, two different shapes:

| Endpoint | Returns | Code |
|---|---|---|
| `GET /api/webhooks/deliveries` | bare `[...]` | `modules/web/misc_routes.py:214-225` |
| `GET /api/deploy/history`      | `{history: [...]}` | `modules/web/settings_routes.py:448-459` |

Their respective frontends:

| Frontend | Expects | Code |
|---|---|---|
| `static/js/settings-notifications.js` | bare list (`Array.isArray(data)` only) | `:109` |
| `static/js/settings-deploy.js` (post-v2.4.12) | both shapes, defensively | `:184-201` |

The deploy panel was originally written to the bare-list convention too — the pre-v2.4.12 handler also only had an `Array.isArray` check, which is what made #137 manifest as *"unexpected response"* in the first place. The backend was the outlier; the frontend got patched defensively, but the asymmetry remains.

## Proposed fix

One-line change in `modules/web/settings_routes.py:459`:

```diff
-            return jsonify({'history': history})
+            return jsonify(history)
```

This aligns `/api/deploy/history` with `/api/webhooks/deliveries` and matches what the rest of the codebase (and the original frontend) already assumed.

## Backward compatibility

Safe. The frontend changed in v2.4.12 explicitly accepts both shapes (`Array.isArray(res.body) || Array.isArray(res.body.history)`), so flipping the backend to a bare list is a no-op for the UI. The comment in the JS handler even names this as the expected migration path.

## Tests

A 4-case contract test already exists on the `fix/issue-137-deploy-history` branch of my fork at `tests/test_issue137_deploy_history.py`:

- `test_deploy_history_returns_bare_array` — populated list, ordered.
- `test_deploy_history_empty_returns_empty_array_not_envelope` — first-run case (the most visible symptom of #137).
- `test_deploy_history_forwards_limit_and_domain_filters` — `?limit=` and `?domain=` still parsed, 200-cap still enforced.
- `test_deploy_history_503_when_deploy_manager_missing` — error path still returns `{"error": "..."}` so the frontend's catch branch can surface the real reason.

Happy to reopen the PR against this issue with those tests included.

## Non-goals

- No frontend change. The v2.4.12 dual-shape handler stays as-is (defensive, forward/backward compatible).
- No other endpoint touched. This is strictly the `/api/deploy/history` alignment.

imagreez:
Closes #152. Follow-up to #137 / #142, per the maintainer's ask in [pull/142#issuecomment-4430183618](https://github.com/fabriziosalmi/certmate/pull/142#issuecomment-4430183618).

## Summary

`GET /api/deploy/history` returned `{"history": [...]}` while the sibling event-log endpoint `GET /api/webhooks/deliveries` (`modules/web/misc_routes.py`) already returned a bare `[...]`. The matching frontend `static/js/settings-notifications.js` reads its sibling with `Array.isArray(data)` directly — that's the established convention for delivery/history endpoints in this codebase. The deploy panel was originally written to it too; v2.4.12 added a dual-shape handler so the panel kept working while a backend flip was pending. This PR completes that flip.

One functional line of change:

```diff
-            return jsonify({'history': history})
+            return jsonify(history)
```

Plus a short docstring callout explaining *why* it's a bare list, so a future maintainer doesn't envelope it back and silently re-introduce the asymmetry.

## Backward compatibility

Safe. The v2.4.12 frontend handler (`static/js/settings-deploy.js:184-201`) is explicitly dual-shape:

```javascript
if (Array.isArray(res.body)) {
    entries = res.body;
} else if (Array.isArray(res.body.history)) {
    entries = res.body.history;
}
```

and its inline comment names this flip as the expected migration path. The error path keeps the `{"error": "..."}` envelope so the frontend's catch branch can still surface a real reason instead of the generic `'unexpected response'` literal.

## Tests

`tests/test_issue152_deploy_history.py` — four cases, all pass, pinning the post-#152 contract:

- `test_deploy_history_returns_bare_array` — populated history comes back as a list, entries in order.
- `test_deploy_history_empty_returns_empty_array_not_envelope` — empty history returns `[]`, not `{"history": []}` (the original visible symptom of #137).
- `test_deploy_history_forwards_limit_and_domain_filters` — `?limit=N` (capped at 200) and `?domain=...` are still parsed and forwarded to `DeployManager.get_history` verbatim. Shape change does not regress the query-string contract.
- `test_deploy_history_503_when_deploy_manager_missing` — error path keeps the `{"error": "..."}` envelope.

Existing `tests/test_deployer.py` (35 cases on `DeployManager.get_history` directly) still passes — the underlying function's return shape was always a list, only the HTTP envelope changed.

## Verification

- [x] `pytest tests/test_issue152_deploy_history.py tests/test_deployer.py -v` → 39/39 pass.
- [ ] Manual: Settings → Deploy → Recent Executions on a fresh install (empty history) renders the empty state, not the red error toast. With one or more deploy events, entries show newest-first.

## Non-goals

- No frontend change. The v2.4.12 dual-shape handler stays as-is (defensive, forward/backward compatible).
- No other endpoint touched. Strictly the `/api/deploy/history` alignment per #152.

### DIFF
--- a/modules/web/settings_routes.py
+++ b/modules/web/settings_routes.py
@@ -448,7 +448,15 @@ def register_settings_routes(app, managers, require_web_auth, auth_manager,
     @app.route('/api/deploy/history', methods=['GET'])
     @auth_manager_ref.require_role('admin')
     def api_deploy_history():
-        """Get deploy hook execution history"""
+        """Get deploy hook execution history.
+
+        Returns a bare list ``[...]`` — matches the sibling event-log
+        endpoint ``/api/webhooks/deliveries`` (modules/web/misc_routes.py)
+        and the convention the UI (``static/js/settings-deploy.js``,
+        ``static/js/settings-notifications.js``) was originally written
+        against. The error path keeps the ``{"error": ...}`` envelope so
+        the frontend's catch branch can surface a real reason.
+        """
         if not deploy_manager:
             return jsonify({'error': 'Deploy manager not available'}), 503
 
@@ -456,7 +464,7 @@ def register_settings_routes(app, managers, require_web_auth, auth_manager,
             limit = min(int(request.args.get('limit', 50)), 200)
             domain = request.args.get('domain')
             history = deploy_manager.get_history(limit=limit, domain=domain)
-            return jsonify({'history': history})
+            return jsonify(history)
         except Exception as e:
             logger.error(f"Failed to get deploy history: {e}")
             return jsonify({'error': 'Failed to get deploy history'}), 500
--- a/tests/test_issue152_deploy_history.py
+++ b/tests/test_issue152_deploy_history.py
@@ -1,0 +1,142 @@
+"""Contract test for issue #152 — align ``GET /api/deploy/history`` with
+the bare-list convention already used by ``GET /api/webhooks/deliveries``
+(``modules/web/misc_routes.py``).
+
+The two endpoints serve the same role for their respective UI panels
+(Settings → Deploy → "Recent Executions" and Settings → Notifications →
+"Recent Deliveries"), yet they returned different shapes: the deliveries
+endpoint returned a bare list, the deploy history endpoint wrapped it in
+``{"history": [...]}``. The original frontend was written to the bare-list
+convention; the v2.4.12 hotfix made it accept both shapes defensively
+(see ``static/js/settings-deploy.js``) so flipping the backend to a bare
+list is a no-op for the UI but removes the asymmetry the maintainer
+flagged in pull/142#issuecomment-4430183618.
+
+These four cases pin the post-#152 contract so a future maintainer
+doesn't "tidy" the bare list back into an envelope and silently
+re-introduce the divergence.
+"""
+
+from unittest.mock import MagicMock
+
+import pytest
+from flask import Flask
+
+from modules.web.settings_routes import register_settings_routes
+
+
+pytestmark = [pytest.mark.unit]
+
+
+def _passthrough_decorator(_min_role):
+    """Stand in for ``auth_manager.require_role(...)`` so the view runs
+    without a real session."""
+    def deco(fn):
+        return fn
+    return deco
+
+
+def _build_app(history_entries):
+    app = Flask(__name__)
+    app.secret_key = 'test'
+
+    auth_manager = MagicMock()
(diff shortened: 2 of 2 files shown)


fabriziosalmi (approved):
Chirurgical. One functional line of diff, a clear docstring explaining *why* the bare list is the contract so a future maintainer doesn't envelope it back, and four targeted tests that pin exactly the right things:

- the populated success case
- the empty-history case (the original visible symptom of #137 and the one most likely to regress)
- the query-string forwarding contract — `?limit=N` capped at 200 and `?domain=...` still threaded through to `DeployManager.get_history(limit=N, domain=...)`
- the error path keeps the `{"error": ...}` envelope so the frontend's catch branch reads the same key as before

Verified locally with the four-case suite running against the patched `settings_routes.py` against the v2.4.12 frontend that already accepts both shapes (`settings-deploy.js:184-201`):

```
$ pytest tests/test_issue152_deploy_history.py -v
tests/test_issue152_deploy_history.py::test_deploy_history_returns_bare_array PASSED
tests/test_issue152_deploy_history.py::test_deploy_history_empty_returns_empty_array_not_envelope PASSED
tests/test_issue152_deploy_history.py::test_deploy_history_forwards_limit_and_domain_filters PASSED
tests/test_issue152_deploy_history.py::test_deploy_history_503_when_deploy_manager_missing PASSED
4 passed
```

Plus the full project unit-suite (140 tests) stays green with this PR applied.

Thanks for the follow-through — picking up the convention point I'd left as a follow-up after #142, opening the issue first, and writing the contract test as the receipt that it stays fixed is exactly the loop I was hoping for. Merging.

PR #154, opened 2026-05-12
fix: APScheduler app_context for background renewal jobs

imagreez:
## Summary

APScheduler background jobs run in worker threads where Flask's thread-local `current_app` proxy is unbound. Every renewal cycle, cron-triggered cleanup, and digest job that tried to access `current_app.config` failed with:

```
RuntimeError: Working outside of application context.
```

Symptoms in the wild were intermittent renewal failures that didn't reproduce from the UI.

## Fix

- Keeps a module-level `_flask_app` reference in `factory.py`
- Extracts a shared `_run_manager_job(manager_key, method_name)` helper that pushes `app.app_context()` before dispatching
- All three scheduler wrappers (`_certificate_renewal_job`, `_client_certificate_renewal_job`, `_weekly_digest_job`) delegate to the helper
- The helper includes defensive logging for edge cases (app not yet initialised, manager not found, unhandled exception)
- `create_app()` sets `_flask_app = app` before `setup_scheduler()` so recovered misfired jobs from the SQLite jobstore can push the context immediately

## Behaviour change

None. Existing installs see no change — if a job was already working without hitting `current_app`, the `app_context()` push is transparent.

## Files

- `modules/core/factory.py`

## Test plan

- [ ] `pytest tests/ -m unit -q` — all pass
- [ ] Restart scheduler with `RENEWAL_THRESHOLD_DAYS=999` so every cert is "due"; confirm no `RuntimeError` in logs

### DIFF
--- a/modules/core/factory.py
+++ b/modules/core/factory.py
@@ -33,6 +33,11 @@ from modules.web import register_web_routes
 logger = get_certmate_logger('factory')
 request_logger = get_certmate_logger('request-watchdog')
 
+# APScheduler jobs run in background threads where Flask's thread-local
+# current_app proxy is unbound.  We keep a module-level reference so we
+# can explicitly push an app context inside those jobs.
+_flask_app = None
+
 
 class AppContainer:
     """DI Container holding all managers and application state"""
@@ -458,28 +463,53 @@ def initialize_managers(container: AppContainer, app):
     }
 
 
+def _run_manager_job(manager_key: str, method_name: str):
+    """Execute a manager method inside a Flask app context.
+
+    APScheduler jobs run in background threads where the thread-local
+    `current_app` proxy is unbound.  We keep a module-level reference to
+    the app instance so we can push an explicit app context before
+    touching any Flask-bound code.
+    """
+    if _flask_app is None:
+        logger.warning(
+            "Background job skipped: Flask app not yet initialised "
+            "(manager_key=%s, method=%s)",
+            manager_key,
+            method_name,
+        )
+        return
+    with _flask_app.app_context():
+        managers = _flask_app.config.get('MANAGERS')
+        manager = managers.get(manager_key) if managers else None
+        if manager is None:
+            logger.warning(
+                "Background job skipped: manager '%s' not found", manager_key
+            )
+            return
+        try:
+            getattr(manager, method_name)()
+        except Exception:
+            logger.exception(
+                "Background job failed: manager_key=%s, method=%s",
+                manager_key,
+                method_name,
+            )
+
+
 def _certificate_renewal_job():
     """Picklable wrapper for certificate renewal check"""
-    from flask import current_app
-    managers = current_app.config.get('MANAGERS')
-    if managers and 'certificates' in managers:
-        managers['certificates'].check_renewals()
+    _run_manager_job('certificates', 'check_renewals')
 
 
 def _client_certificate_renewal_job():
     """Picklable wrapper for client certificate renewal check"""
-    from flask import current_app
-    managers = current_app.config.get('MANAGERS')
-    if managers and 'client_certificates' in managers:
-        managers['client_certificates'].check_renewals()
+    _run_manager_job('client_certificates', 'check_renewals')
 
 
 def _weekly_digest_job():
     """Picklable wrapper for weekly digest"""
-    from flask import current_app
-    managers = current_app.config.get('MANAGERS')
-    if managers and 'digest' in managers:
-        managers['digest'].send()
+    _run_manager_job('digest', 'send')
 
 
 def setup_scheduler(container: AppContainer):
@@ -768,12 +798,17 @@ def create_app(test_config=None):
 
     configure_app(container, app, test_config)
     initialize_managers(container, app)
+    app.config['MANAGERS'] = container.managers
     setup_api(container, app)
     register_web_routes(app, container.managers)
     setup_csrf_protection(app)
     setup_security_headers(app)
     setup_rate_limiting(app, container)
     setup_slow_request_logging(app, container)
+
+    # Make the app instance available to background APScheduler jobs before
+    # starting the scheduler so recovered misfired jobs can push an app context.
+    _flask_app = app
     setup_scheduler(container)
 
     return app, container
--- a/modules/core/factory.py
+++ b/modules/core/factory.py
@@ -473,10 +473,9 @@ def _run_manager_job(manager_key: str, method_name: str):
     """
     if _flask_app is None:
         logger.warning(
-            "Background job skipped: Flask app not yet initialised "
-            "(manager_key=%s, method=%s)",
-            manager_key,
-            method_name,
+            "Background job skipped: Flask app not yet initialised",
+            manager_key=manager_key,
+            method_name=method_name,
         )
         return
     from flask import current_app
@@ -485,16 +484,17 @@ def _run_manager_job(manager_key: str, method_name: str):
         manager = managers.get(manager_key) if managers else None
         if manager is None:
             logger.warning(
-                "Background job skipped: manager '%s' not found", manager_key
+                "Background job skipped: manager not found",
+                manager_key=manager_key,
             )
             return
         try:
             getattr(manager, method_name)()
         except Exception:
             logger.exception(
-                "Background job failed: manager_key=%s, method=%s",
-                manager_key,
-                method_name,
+                "Background job failed",
+                manager_key=manager_key,
+                method_name=method_name,
             )
 
 
--- a/modules/core/factory.py
+++ b/modules/core/factory.py
@@ -479,8 +479,9 @@ def _run_manager_job(manager_key: str, method_name: str):
             method_name,
         )
         return
+    from flask import current_app
     with _flask_app.app_context():
-        managers = _flask_app.config.get('MANAGERS')
+        managers = current_app.config.get('MANAGERS')
         manager = managers.get(manager_key) if managers else None
         if manager is None:
             logger.warning(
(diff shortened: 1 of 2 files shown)


fabriziosalmi (changes requested):
The split + the design is clean — three picklable wrappers collapsed onto one helper that does the context push, the MANAGERS bridge through `app.config` is the right plumbing, the defensive logging branches (app not initialised, manager missing, exception) all make sense. Thanks for breaking #141 apart as discussed.

One bug, but it's a load-bearing one: the new `_flask_app = app` assignment inside `create_app()` is missing the `global _flask_app` declaration, so it creates a **local** variable that shadows the module-level one. The module-level `_flask_app` stays `None`, the helper hits its first guard, and every renewal / client-renewal / digest job is silently skipped. The PR makes the situation worse than before — at least pre-PR the jobs ran (with a broken `current_app`); after the merge they don't run at all.

Reproduction in three lines:

```python
>>> _x = None
>>> def setup():
...     _x = 'app'
...
>>> setup(); print(_x)
None
```

Fix is one line at the top of `create_app()`:

```python
def create_app(test_config=None):
    global _flask_app                          # <-- add this
    ...
    _flask_app = app
    setup_scheduler(container)
```

That said, please **also** add a regression test before re-requesting review, so a future refactor doesn't silently re-introduce this — the unit suite never exercises the scheduler path, and the manual test in the PR description ("Restart scheduler with RENEWAL_THRESHOLD_DAYS=999") is the only thing that would currently catch it. Something like:

```python
def test_create_app_populates_module_level_flask_app(monkeypatch, tmp_path):
    monkeypatch.setenv('DATA_DIR', str(tmp_path))
    from modules.core import factory
    factory._flask_app = None  # known clean state
    app, _ = factory.create_app(test_config={'TESTING': True})
    assert factory._flask_app is app
```

Two non-blocking observations while we're at it:

1. `_run_manager_job` reads `managers = _flask_app.config.get('MANAGERS')` after pushing the context — at that point you can equivalently use the standard `from flask import current_app; managers = current_app.config.get('MANAGERS')` and avoid relying on `_flask_app.config` directly. Minor stylistic point; whichever you prefer is fine.

2. The misfire-recovery comment ("recovered misfired jobs from the SQLite jobstore can push the context immediately") is accurate but it surfaces a subtle ordering invariant. Once the global is fixed, consider adding a one-line sanity assertion at the top of `setup_scheduler` so a future caller can't accidentally call it with `_flask_app is None`:

   ```python
   assert _flask_app is not None, "setup_scheduler called before _flask_app was set"
   ```

Once the `global` keyword + a regression test are in, happy to LGTM and merge.

fabriziosalmi:
Heads-up while you're considering the `global` fix above: your other two PRs from the #141 split (#153 and #155) just shipped in [v2.4.17](https://github.com/fabriziosalmi/certmate/releases/tag/v2.4.17) — the bare-list deploy history (closing #152) and the EXTRA_REQUIREMENTS Dockerfile build-arg + `certbot-dns-azure` baked-in. Both went in clean, thanks for the split.

No pressure on this one — the `global _flask_app` fix + a regression test is the only thing in the way. Come back to it when it makes sense; the PR will be here.

imagreez:
Hi @fabriziosalmi 

I'll try to have the changes ready within the next hour!

imagreez:
Addressed the review feedback as discussed:

1. Added  at the top of  so the assignment  updates the module-level variable instead of creating a local that shadows it.
2. Replaced  with  inside  since an app context is already active there.
3. Added a sanity assertion at the top of  —  — so future callers can't accidentally invoke it before the global reference is populated.
4. Added a regression test  in  so a future refactor can't silently re-introduce this invariant failure.

All green locally. Ready for another look — thanks again!

fabriziosalmi (commented):
Thanks — all four points from the previous review landed exactly as discussed:

1. `global _flask_app` at the top of `create_app()` ✓
2. `current_app.config.get('MANAGERS')` (the stylistic nit) ✓
3. `assert _flask_app is not None` in `setup_scheduler` ✓
4. `tests/test_factory_scheduler.py` regression test ✓

Verified locally:

```
$ pytest tests/test_factory_scheduler.py -v
tests/test_factory_scheduler.py::test_create_app_populates_module_level_flask_app PASSED
$ pytest tests/ -q   # full unit suite + this PR
141 passed in 4.95s
```

And a smoke test confirms the happy path dispatches correctly:

```python
factory._flask_app = None
app, _ = factory.create_app(test_config={'TESTING': True})
assert factory._flask_app is app                          # module-level populated
app.config['MANAGERS'] = mocks
factory._certificate_renewal_job()                        # dispatches OK
factory._client_certificate_renewal_job()                 # dispatches OK
factory._weekly_digest_job()                              # dispatches OK
```

So the core fix is solid and merge-ready on the happy path.

## One follow-up while I was poking: latent crash in the three guard branches

Running the same smoke test against the failure paths surfaces a second issue — independent of the original `global` bug, easy to miss because the guards are exactly the branches you don't see during normal startup:

```
Branch 1 (_flask_app is None at misfire-recovery): CRASH on logger —
    StructuredLogger.warning() takes 2 positional arguments but 4 were given
Branch 2 (manager not in MANAGERS):                 CRASH on logger —
    StructuredLogger.warning() takes 2 positional arguments but 3 were given
Branch 3 (manager.check_renewals() raises):         CRASH on logger —
    StructuredLogger.exception() takes 2 positional arguments but 4 were given
```

Cause: `modules.core.structured_logging.StructuredLogger.warning` / `.exception` signatures are `(self, msg: str, **kwargs)` — keyword-only, no positional `*args` for `%s` interpolation. The three `logger.*` calls inside `_run_manager_job` use the stdlib `%s` style, so the moment a guard actually fires, the helper raises `TypeError` and APScheduler logs that instead of the real reason for skipping. Branch 3 is the practical concern in production: when a renewal genuinely fails (ACME timeout, DNS prop, plugin error), the original `RuntimeError`/etc. is masked by a `TypeError` from the logger call meant to *record* the failure.

Fix is mechanical — match the project's existing structured-logging convention (e.g. the `request_logger.warning("Slow request completed", method=..., path=..., ...)` block in the slow-request watchdog right above this code):

```python
def _run_manager_job(manager_key: str, method_name: str):
    if _flask_app is None:
        logger.warning(
            "Background job skipped: Flask app not yet initialised",
            manager_key=manager_key,
            method_name=method_name,
        )
        return
    from flask import current_app
    with _flask_app.app_context():
        managers = current_app.config.get('MANAGERS')
        manager = managers.get(manager_key) if managers else None
        if manager is None:
            logger.warning(
                "Background job skipped: manager not found",
                manager_key=manager_key,
            )
            return
        try:
            getattr(manager, method_name)()
        except Exception:
            logger.exception(
                "Background job failed",
                manager_key=manager_key,
                method_name=method_name,
            )
```

Plus a couple more assertions in `tests/test_factory_scheduler.py` so this exact class of bug doesn't re-enter via a future refactor:

```python
def test_run_manager_job_handles_uninitialised_app_without_crash():
    factory._flask_app = None
    # Must not raise — silent skip is the contract
    factory._run_manager_job('certificates', 'check_renewals')

def test_run_manager_job_handles_missing_manager_without_crash(monkeypatch, tmp_path):
    monkeypatch.setenv('DATA_DIR', str(tmp_path))
    factory._flask_app = None
    app, _ = factory.create_app(test_config={'TESTING': True})
    app.config['MANAGERS'] = {}
    factory._run_manager_job('certificates', 'check_renewals')

def test_run_manager_job_swallows_manager_exception_without_crash(monkeypatch, tmp_path):
    monkeypatch.setenv('DATA_DIR', str(tmp_path))
    factory._flask_app = None
    app, _ = factory.create_app(test_config={'TESTING': True})
    class Boom:
        def check_renewals(self): raise RuntimeError('simulated')
    app.config['MANAGERS'] = {'certificates': Boom()}
    factory._run_manager_job('certificates', 'check_renewals')
```

## How you want to handle it

Two equally-fine paths, your call — both honour your time:

**a)** push one more commit with the kwargs flip + the three regression tests, then I merge. ~10 min of your time.

**b)** I merge as-is now (the happy path is correct and that's what 99% of runs hit) and immediately open a follow-up PR myself with the logger fix + tests, crediting you in the body. The latent crash never reaches production users this way either, and you don't have to do a third round.

Either way the original bug is fixed and the new behaviour is strictly better than pre-PR. Let me know which you prefer; in the meantime the work you've done here is solid and appreciated.

imagreez:
PR is clean and ready for final review.

**What changed since the last round:**

1. **StructuredLogger kwargs fix** (76e3a21) — The three guard branches in `_run_manager_job` now use keyword arguments instead of positional `%s` interpolation, matching `StructuredLogger.warning/exception` signatures. Previously these branches would themselves raise `TypeError` if ever hit.

2. **Guard-branch regression tests** — `test_factory_scheduler.py` now covers:
   - App not yet initialised (`_flask_app = None`)
   - Manager missing from `MANAGERS`
   - Manager method raising an exception
   All three assert silent skip / no crash.

**Verification:**
```
$ pytest tests/test_factory_scheduler.py -v
4 passed

$ pytest tests/ -q
393 passed, 10 skipped
```

Full unit-suite green. Merge when ready.

fabriziosalmi (approved):
LGTM — option (a) landed exactly as discussed. Verified locally:

**pytest**
- `tests/test_factory_scheduler.py`: 4/4 PASS (population + 3 guard branches)
- Full unit suite: 52/52 PASS, no regressions.

- Boot: `Scheduler started` + 3 jobs added to job store; `/health` reports `scheduler: running`.
- Force-fire happy path on real managers: no exception.
- Force-fire guard branch (`_flask_app is None` in a separate `docker exec python` process): logged `Background job skipped: Flask app not yet initialised` 4× — kwargs flip on `logger.warning`/`logger.exception` works, no TypeError. The exact latent crash flagged in the second review is confirmed gone.

Merging.

PR #155, opened 2026-05-12
build: EXTRA_REQUIREMENTS build-arg + certbot-dns-azure baked in

imagreez:
## Summary

Two build/deploy improvements:

1. **`EXTRA_REQUIREMENTS` build-arg** — lets a single `docker build` layer in optional storage backends (Azure KV, AWS SM, Vault, Infisical) without forking the image.
2. **`certbot-dns-azure==2.5.0`** baked into the main `requirements.txt` — Azure DNS is one of the four major providers already in the UI but picking it produced "plugin not installed" errors on default builds.

### `EXTRA_REQUIREMENTS` build-arg

Accepts a space-separated list of requirements file paths and runs `pip install -r <file>` for each, after the main `REQUIREMENTS_FILE` install:

```bash
docker build \
  --build-arg EXTRA_REQUIREMENTS="requirements-azure-storage.txt requirements-aws-storage.txt" \
  -t certmate:azure+aws .
```

Defaults to empty, so existing builds are unchanged. Wired through `docker-compose.yml` via env var. The Dockerfile COPY layer was widened to `requirements*.txt` so any optional file is available for pip without rebuild.

### `certbot-dns-azure`

Pinned to 2.5.0 — 2.6.0 jumped to `certbot>=3.0`, which would conflict with the repo-wide `certbot==2.10.0` pin.

## Files

- `Dockerfile`, `docker-compose.yml`, `requirements.txt`

## Test plan

- [ ] `docker build` succeeds with and without `EXTRA_REQUIREMENTS`
- [ ] `docker build --build-arg EXTRA_REQUIREMENTS=requirements-azure-storage.txt` includes Azure KV packages
- [ ] Container boots, selecting "Azure" as DNS provider no longer produces "plugin not installed"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,10 @@ RUN apt-get update && \
     apt-get install -y -o Acquire::Retries=3 gcc && \
     rm -rf /var/lib/apt/lists/*
 
-# Copy requirements first for better caching
-COPY requirements.txt requirements-minimal.txt ./
+# Copy every requirements*.txt so REQUIREMENTS_FILE and EXTRA_REQUIREMENTS
+# can point at any of the optional sets (storage backends, cloud DNS,
+# extended providers, …) without rebuilding the COPY layer for each one.
+COPY requirements*.txt ./
 
 # Create virtual environment and install dependencies
 RUN python -m venv /opt/venv
@@ -19,8 +21,17 @@ ENV PATH="/opt/venv/bin:$PATH"
 # Install minimal requirements by default (fastest build)
 # Override with --build-arg REQUIREMENTS_FILE=requirements.txt for full install
 ARG REQUIREMENTS_FILE=requirements.txt
+# Optional second pip install for storage backends or extra DNS plugins.
+# Examples:
+#   --build-arg EXTRA_REQUIREMENTS=requirements-storage-all.txt
+#   --build-arg EXTRA_REQUIREMENTS=requirements-azure-storage.txt
+# Empty by default → no second install, layer cached.
+ARG EXTRA_REQUIREMENTS=
 RUN pip install -U pip setuptools wheel && \
-    pip install --no-cache-dir -r ${REQUIREMENTS_FILE}
+    pip install --no-cache-dir -r ${REQUIREMENTS_FILE} && \
+    if [ -n "${EXTRA_REQUIREMENTS}" ]; then \
+        pip install --no-cache-dir -r ${EXTRA_REQUIREMENTS}; \
+    fi
 
 # Production stage
 FROM python:3.12-slim-trixie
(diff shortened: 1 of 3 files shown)


fabriziosalmi (approved):
Clean and well-thought-out. Two non-blocking observations after a careful read.

## What the PR gets right

**`EXTRA_REQUIREMENTS` build-arg.** Space-separated list, defaults to empty, intentional word-splitting with a `shellcheck disable=SC2086` annotation so a future maintainer doesn't "fix" the quoting and break the loop. The `COPY requirements*.txt ./` glob means any optional set is available without rebuilding the COPY layer per combination. The compose wiring uses `${EXTRA_REQUIREMENTS:-}` so the env var is genuinely optional and existing `.env` files don't need to change.

The pattern is what the optional-storage-backend ecosystem in this repo has been waiting for — `requirements-azure-storage.txt`, `requirements-aws-storage.txt`, `requirements-vault-storage.txt`, `requirements-infisical-storage.txt` all exist and can now be layered into a single image without forking the Dockerfile.

**`certbot-dns-azure==2.5.0`** in main requirements. The pin reasoning is correct — 2.5.0 is the last in the `certbot<3.0,>=2.0` line; 2.6.0+ pulled the certbot dependency to 3.x and would conflict with the repo-wide `certbot==2.10.0`. The comment block in the diff captures this so a future dependabot bump doesn't silently relax the pin and break the install. (The same constraint already lived in `requirements-azure.txt`'s comment block — the PR is just promoting it.)

This also closes a real UX bug: the UI dropdown lists Azure as a DNS provider, but the default image had no `certbot-dns-azure` plugin and selecting it would fail at certificate creation with "plugin not installed". Baking it in keeps the dropdown's affordance honest.

## Non-blocking notes

**1. Image size delta (~30–50 MB).** `certbot-dns-azure` pulls in `azure-identity`, `azure-mgmt-dns`, `msrest`, `msal`, and their transitive dependencies. The default image grows. The trade-off is defensible — `cloudflare`, `route53`, `digitalocean`, `google` are already baked in with their own SDK weight, and Azure is in the same UX tier — but it's worth documenting the size change in the v2.4.17 release notes so anyone running the image in a constrained environment (edge ARM hosts, etc.) isn't surprised. If the size becomes a complaint, the future move is to split into a `certmate:slim` (DNS plugins as `EXTRA_REQUIREMENTS`-only) and a `certmate:full` (all four), which is exactly what the new build-arg enables.

**2. No build-level CI test.** Dockerfile changes aren't easily unit-testable, but the existing `.github/workflows/docker-multiplatform.yml` could be extended with one extra job that runs `docker build --build-arg EXTRA_REQUIREMENTS=requirements-azure-storage.txt ...` and asserts the resulting image imports `azure.keyvault.secrets`. Not blocking for this PR — flag as a follow-up so the `EXTRA_REQUIREMENTS` contract has a regression test.

LGTM, merging.

PR #156, opened 2026-05-12
feat: configurable cert key type/size (RSA 2048/3072/4096 + ECDSA P-256/P-384)

imagreez:
## Summary

Every cert issued by CertMate was hardcoded to RSA-2048 because `CAManager.build_certbot_command` never passed `--key-type` or `--rsa-key-size` to certbot. Two real-world asks made this awkward:

- **Compliance**: some operators are required to use RSA-3072 / RSA-4096.
- **Modern stacks**: ECDSA produces smaller certs and faster TLS handshakes; many CDNs and load balancers prefer it.

Both groups had to patch the code or maintain a fork.

## What it does

- **Global default** under Settings → General → "Default Certificate Key": `default_key_type` (`rsa` / `ecdsa`), `default_key_size` (`2048` / `3072` / `4096`), `default_elliptic_curve` (`secp256r1` / `secp384r1`). The size and curve selectors are mutually exclusive — picking ECDSA hides the size picker and vice versa.
- **Per-cert override** under the create-cert form's "Advanced Options". Leaving it on "Use global default" sends no key fields, and the backend inherits the configured default.
- **Renewals preserve the original shape automatically**, because certbot persists the `--key-type` / `--rsa-key-size` / `--elliptic-curve` flags into its own `renewal/<domain>.conf` at create time. No new bookkeeping in CertMate.
- **Backwards-compatible default**: settings without these fields migrate to `rsa` / `2048` / `secp256r1` at load time, so existing installs see no behaviour change after upgrade.

### Validation

A `validate_key_options(key_type, key_size, elliptic_curve)` helper (`modules/core/utils.py`) rejects contradictory shapes up-front with a 400:

- `key_type='rsa'` with an `elliptic_curve` override
- `key_type='ecdsa'` with a `key_size` override
- Unsupported sizes (e.g. 1024) or curves (e.g. `secp521r1`)

### Bug fix — Settings dropdown empty on reload

The `<select>` for "Key Type" rendered empty on every page reload because the masking regex `(token|secret|password|key|credential)` matched the `key` substring in `default_key_type`, returning `'********'` which matches no `<option>`. Fixed via an explicit `_NON_SECRET_KEYS` allowlist in the GET settings route.

### v2.4.12 integration

- `default_key_type`, `default_key_size`, `default_elliptic_curve` are added to `PUBLIC_SETTINGS_WRITABLE_KEYS` so the UI `POST /api/settings` saves them.
- Domain scope: the existing `_check_domain_scope(domain, 'create')` in the create-cert endpoint already covers the whole request — the per-cert key override is a *how*, not a *which-domain*, so no new scope gate is needed.

## Files

| File | Purpose |
|------|---------|
| `modules/core/utils.py` | `validate_key_options()` pure-function validator |
| `modules/core/settings.py` | Global defaults, legacy migration, `PUBLIC_SETTINGS_WRITABLE_KEYS`, save-time validation |
| `modules/core/ca_manager.py` | Emit `--key-type`/`--rsa-key-size`/`--elliptic-curve` on certbot CLI |
| `modules/core/certificates.py` | Plumb key shape through `create_certificate()`, fallback to global defaults, defensive `TypeError` path |
| `modules/api/models.py` + `resources.py` | API schema + route for key shape on cert creation and global settings |
| `modules/web/settings_routes.py` | `_NON_SECRET_KEYS` allowlist in GET settings masking |
| `templates/index.html` + `partials/settings_general.html` | UI selectors |
| `static/js/dashboard.js` + `settings.js` | `toggleCertKeyOptions()` / `toggleDefaultKeyOptions()`, populate on load, form submit |
| `tests/test_key_options.py` | 20 test cases: accept paths, reject paths, command builder, settings migration, save-time rejection |
| `tests/test_settings_masking_allowlist.py` | 4 regression tests: allowlisted fields unmasked, secrets masked, nested recursion, empty-string passthrough |

## Test plan

- [ ] `pytest tests/test_key_options.py -v` — 20/20
- [ ] `pytest tests/test_settings_masking_allowlist.py -v` — 4/4
- [ ] `pytest tests/ -m unit -q --ignore=tests/test_secret_key_env.py` — all pass
- [ ] Save ECDSA/secp384r1 as global default; reload Settings page → dropdowns reflect persisted values
- [ ] Issue a cert with ECDSA/P-384 override; `openssl x509 -in fullchain.pem -text` shows `id-ecPublicKey` / `NIST CURVE: P-384`
- [ ] Issue a cert with "Use global default" + global=ECDSA; cert inherits the global shape
- [ ] Renew the cert; certbot reuses the curve (no regression to RSA)


fabriziosalmi:
Acknowledging this is queued and not forgotten. Of the four PRs from the #141 split, two (#153 + #155) shipped in [v2.4.17](https://github.com/fabriziosalmi/certmate/releases/tag/v2.4.17), one (#154) is waiting on a one-line `global` fix, and this is the substantive one that wants a proper review window (the +818 LOC across 15 files, ECDSA support, per-cert override path, 241 lines of tests — none of that should be reviewed in a hurry).

Planning to take this in the next sprint window, probably early next week so you have breathing room between releases. The branch is mergeable as-is so the rebase debt won't pile up. No need to do anything on your side in the meantime.

fabriziosalmi:
Queued for review + merge tonight, in the same window as the broader v3 UI fix pass we have batched. Plan is to run the test plan from the PR body locally (the 20 unit cases plus issuing a real cert with both ECDSA P-384 and RSA-3072 to confirm the certbot CLI carries the flags end-to-end and the renewal preserves the shape) before pressing merge — that's the smoke gate we settled on after the v2.4.17 wave to avoid shipping anything that affects the certbot command line without a real-cert signal.

Thanks for the patience on this one; the v2.4.17 stability/regression queue had to land first.

Issue #158, opened 2026-05-12
Authentication issues after upgrade.


tuxpowered:
**Describe the bug**
Upgrading from 2.4.7 to 2.4.8 caused authentication issues.  downgraded to 2.4.7 authentication restored. Updated this morning to 2.4.16 hoping that the furry of updates fixed the issue but it does not appear to be so.

Downgrading again to 2.4.7 looses all authentication, wizard starts backup, and no certificates exist that previously were present.

**To Reproduce**

**Expected behavior**
When upgrading from one version to another that authentication methods should stay consistant.
On disk all the certs are still there, so certmate should have picked them up, even when rolling back but it completely ignores it. 

**Screenshots**
This is the docker logs from 2.4.16
```
[2026-05-12 14:25:20 +0000] [8] [INFO] Starting gunicorn 23.0.0

[2026-05-12 14:25:20 +0000] [8] [INFO] Listening at: http://0.0.0.0:8000 (8)

[2026-05-12 14:25:20 +0000] [8] [INFO] Using worker: gthread

[2026-05-12 14:25:20 +0000] [9] [INFO] Booting worker with pid: 9

INF Persistent volume verified — data directory survives container restarts | timestamp=2026-05-12T14:25:21.766782Z logger=certmate.factory host=0c103b3e3640 pid=9

INF Updated deployment cache TTL to 300 seconds | timestamp=2026-05-12T14:25:21.769338Z logger=modules.core.cache host=0c103b3e3640 pid=9

INF CA already exists, loading from disk | timestamp=2026-05-12T14:25:21.769625Z logger=modules.core.private_ca host=0c103b3e3640 pid=9

INF CA loaded successfully | timestamp=2026-05-12T14:25:22.115160Z logger=modules.core.private_ca host=0c103b3e3640 pid=9

INF Rate limiting configured with 8 endpoint limits | timestamp=2026-05-12T14:25:22.115645Z logger=modules.core.rate_limit host=0c103b3e3640 pid=9

INF Scheduler started | timestamp=2026-05-12T14:25:22.211471Z logger=apscheduler.scheduler host=0c103b3e3640 pid=9

INF Added job "_certificate_renewal_job" to job store "default" | timestamp=2026-05-12T14:25:22.220298Z logger=apscheduler.scheduler host=0c103b3e3640 pid=9

[...]

PR #160, opened 2026-05-12
Add CertMate password reset tool

tuxpowered:
This script allows resetting a user's password in CertMate, utilizing its authentication module. It includes options for a dry run and creates a backup of the settings file before applying changes.

In the last sprint of updates password authentication seems to have changed, so this should allow a user to reset their password easily, 

backups of the settings.json file are made before any changes are applied.

### DIFF
--- a/modules/core/settings.py
+++ b/modules/core/settings.py
@@ -10,7 +10,12 @@ from pathlib import Path
 
 from .constants import iter_cert_domain_dirs
 from .file_operations import FileOperations
-from .utils import generate_secure_token, validate_email, validate_api_token, validate_domain
+from .utils import (
+    generate_secure_token,
+    validate_email,
+    validate_api_token,
+    validate_domain,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


fabriziosalmi:
Thanks for the contribution and for filing #158 in the first place — the detailed log dump and the docker-compose excerpt were what unblocked the diagnosis. Wanted to give you a heads-up on three things before this goes any further.

## Status of #158 itself

PR #163 (rocogamer) was merged earlier today and addresses the destructive downgrade path that wiped your settings.json. The fix:

- Detects a downgrade on boot (`settings.json` carries a `certmate_version` stamp now) and logs a loud ERROR pointing at the unified backup folder.
- Surfaces a `certmate_recovery_suggested` flag from `GET /api/web/settings` when the wizard would otherwise overwrite an installation that still has certificates on disk; the UI now shows a "Existing Data Detected" modal with a "Restore from Backup" CTA.
- Ships `scripts/reset_admin_password.py` for the emergency-access case you hit. Same shape as your tool: bcrypt the new password, write it back into `settings.json`, keep all other state untouched, take a pre-write backup of the file.

So the script your PR adds is now functionally a duplicate of one that already landed.

## What's puzzling about this PR's diff

`modules/core/settings.py` shows `-260` lines on top of `+318`. Looking at the patch, those removals overlap with code that's still load-bearing on `main` post-#163 (the migration, masking, and version-stamp paths). My guess is the branch was forked from an older base and the rebase pulled in deletions that no longer correspond to current `main`.

Two paths from here, whichever you prefer:

1. **Tool-only rebase.** Drop the `settings.py` changes entirely (or reduce them to additions only — e.g. anything your script genuinely needs that isn't already in `auth.py`), keep `reset_password.py` as a complementary CLI. If your tool has features the in-tree script doesn't (e.g. dry-run, settings backup with a more descriptive name, `--username` flag), those would be a welcome differentiator. The settings-store side is best left alone.

2. **Close as superseded.** Equally fine if you'd rather move on. The admin-reset path is covered by the in-tree script, and #158 itself is closed.

Either way, the original issue you opened got fixed because you opened it. Thanks again.

fabriziosalmi:
Thanks for the contribution and apologies for the 8-day silence — the audit + v2.6.x release wave swallowed the queue. Going through it now.

## What works in your proposal

Your tool has design improvements over what currently lives in `scripts/reset_admin_password.py`:

- `--username` parameter (arbitrary user, not just `admin`)
- `--dry-run` to print the resulting hash without writing
- `AuthManager`-based hashing rather than calling `bcrypt` directly — keeps the tool in sync with whatever password scheme the project moves to
- Pre-write backup of `settings.json` (good)

These are all real upgrades and I would happily merge them.

## What is blocking the merge

The PR as it stands is unreviewable because of unrelated changes that drown the actual feature:

1. **`modules/core/settings.py` (+437/-259) is ~97% formatting churn** — single-quote → double-quote conversion, multi-line tuple expansion, parenthesisation. Likely a stray `black` run. The real functional change is **two lines**: adding `storage_backend` and `default_ca` to `PUBLIC_SETTINGS_WRITABLE_KEYS`.

2. **Both of those additions are problematic against current `main`**:
   - `default_ca` is **already** in `PUBLIC_SETTINGS_WRITABLE_KEYS` (line 14 of the current file). The `# ADD THIS` is leftover from an older base.
   - `storage_backend` should NOT be added. The canonical key for storage configuration is `certificate_storage` (a dict carrying the per-backend config). Recent PRs (#214/#215/#216) all use it. Introducing `storage_backend` as a sibling top-level key creates a parallel surface that conflicts with the dedicated `POST /api/storage/config` endpoint.

3. **DIRTY merge state** — the formatting churn conflicts with the seven audit-fix PRs that landed today (#219–#225) on `settings.py`.

## The path forward

I am going to close this PR rather than asking you to rebase the current branch — the formatting churn will keep reconflicting and resolving merge conflicts on cosmetic changes wastes everyone's time.

**If you can spare another 30 minutes**, I would love a clean re-submission with ONLY the tool itself:

- New file `scripts/reset_user_password.py` (note: alongside the existing `scripts/reset_admin_password.py`, not replacing it — different audience).
- Optional: a short README note in the `scripts/` folder explaining the difference between the two (`reset_admin_password.py` = quick admin-only emergency reset, `reset_user_password.py` = generic with `--username` / `--dry-run` / `AuthManager`-based hashing / `settings.json` backup).
- **No changes to `modules/core/settings.py`** — leave that file untouched.
- **No changes to `static/js/settings.js`** — same reason.

That PR would be a fast review and a quick merge. Thanks again for the work.

— Fab

PR #161, opened 2026-05-12
Update Docker Compose configuration - allow enabling of proxy handling

ITJamie:
currently theres no way with the docker compose setup to handle setting of proxy headers (there is code but no way to enable it)
thus all user sessions get marked as from the same ip when coming from nginx

this causes rate limiting

```
certmate  | {"timestamp": "2026-05-12T15:48:49.161057Z", "level": "warning", "logger": "modules.core.rate_limit", "message": "Rate limit exceeded for 172.18.0.1 on certificate_list", "source": {"file": "rate_limit.py", "line": 122, "function": "is_allowed"}, "host": "d28e618fafef", "pid": 9}
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@ services:
       - CLOUDFLARE_TOKEN=${CLOUDFLARE_TOKEN}
       - API_BEARER_TOKEN=${API_BEARER_TOKEN:-}  # Auto-generated if empty; set a strong value for production
       #- API_BEARER_TOKEN_FILE=${API_BEARER_TOKEN_FILE:-}  # Alternative: path to a file containing the bearer token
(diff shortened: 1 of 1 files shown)


fabriziosalmi (approved):
LGTM. The 1-liner exposes an env var that's already consumed by the application — see [modules/core/factory.py:342-344](https://github.com/fabriziosalmi/certmate/blob/main/modules/core/factory.py#L342-L344), where setting `BEHIND_PROXY` to `true`/`1`/`yes` wraps `app.wsgi_app` with `ProxyFix(x_for=1, x_proto=1, x_host=1, x_prefix=1)`. Without this line in `docker-compose.yml`, operators running behind nginx or another reverse proxy have no obvious surface to flip — every session is attributed to the proxy's IP and the per-IP rate limiter trips collectively, which is exactly what the warning in your bug report ("Rate limit exceeded for 172.18.0.1 on certificate_list") shows.

Default `false` keeps existing setups unchanged. Merging.

PR #206, opened 2026-05-18
feat: v2.7.0 - OIDC/SSO authentication (close #191)

### LINKED ISSUE #191: [Feature] OIDC/SSO support
No.

I would like to add an OIDC/OAuth2/SSO identity source, handle filter option for groups. The identity provider would make a callback and let the provider generate the proper token to authenticate and authorize the user

No.

**Additional context**
Many other tool has proper settings for it. I don't want to give you full implementation details.
BTW, I will separately create another feature request for LDAP/AD.

imagreez:
## Summary

Adds OIDC/OAuth2/SSO support as an **additive** identity source alongside local auth and API keys. Closes #191.

- **Flow**: Authorization Code + PKCE via [Authlib](https://docs.authlib.org/) (signature/audience/issuer/expiry/nonce all validated by the library against the IdP's JWKS).
- **Provisioning**: JIT — any IdP-authenticated user gets a `users` row on first login. `password_hash=''` so SSO-only accounts can't fall back to local login.
- **Role mapping**: admin-configurable arbitrary claim (default `groups`) + `{claim_value → role}` table; first match wins. Configurable default role when no rule matches (default `viewer`).
- **Linking**: when `link_by_email` is on, an IdP login that matches an existing local user's email is linked instead of duplicated; the existing role is preserved.
- **Config UI**: new **Settings → SSO** tab. `client_secret` masked (`********`) on GET and round-trip-safe on POST.
- **Logout**: returns `oidc_logout_url` for OIDC-minted sessions when the IdP exposes `end_session_endpoint` and a `post_logout_redirect_uri` is configured.
- **Decorators untouched**: `@require_auth` / `@require_role` cannot tell which source authenticated the user. A new `source` kwarg on `AuthManager.create_session` is metadata only.

### Files

- New: `modules/core/oidc.py`, `modules/web/oidc_routes.py`, `templates/partials/settings_oidc.html`, `static/js/settings-oidc.js`, `tests/test_oidc_{manager,routes,settings_masking}.py`
- Modified: `modules/{__init__,core/{auth,factory,settings},web/{auth_routes,routes}}.py`, `templates/{login,settings}.html`, `requirements.txt`, `README.md`

### Security

- PKCE (S256) forced for every flow regardless of client type.
- `state` + `nonce` validated by Authlib; `code_verifier` stashed server-side in `flask.session` (signed by `app.secret_key`).
- `oidc` is on `SETTINGS_REJECT_KEYS` — the bulk `/api/web/settings` POST cannot mutate it; only the dedicated `/api/auth/oidc/settings` endpoint (admin-only, audited) can.
- Callback failures count against the existing per-IP login rate limit.
- Open-redirect guard on `?next=` (same-origin absolute paths only).

### Backwards compatibility

- Default: `oidc.enabled = false`. Behavior unchanged for existing installs.
- `essential_keys` backfill ensures upgraded installs receive the default `oidc` block on first load.
- Local users, API keys and the legacy bearer token continue working.

## Test plan

- [x] 36 new unit tests pass (`tests/test_oidc_manager.py`, `tests/test_oidc_routes.py`, `tests/test_oidc_settings_masking.py`)
- [x] Full unit + integration suite: **477 passed, 0 failed**
- [x] App boots cleanly with `oidc` in `container.managers`
- [x] All 5 routes (`/api/auth/oidc/{config,login,callback,settings GET/POST}`) registered on the Flask app
- [x] Manual smoke test against a Keycloak/Authentik instance (recommended before merge): configure redirect URI `<base>/api/auth/oidc/callback`, complete login, verify role mapping + session cookie continuity for protected pages
- [x] Downgrade safety: setting `oidc.enabled=true` then running a 2.6.x process correctly ignores the block (warning logged by existing downgrade detector)


fabriziosalmi (changes requested):
Thanks for the substantial work here. Reviewed at commit 928c4c8.

## Playbook results

Applied the rigorous-review playbook end-to-end against the PR head:
- Full pytest on the worktree: **901 passed, 14 skipped, 2 xfailed, 0 failed** in 146s. The 36 new OIDC tests pass cleanly.
- Diff review of `modules/core/oidc.py` (534), `modules/web/oidc_routes.py` (235), the diffs to `auth.py` / `factory.py` / `settings.py` / `auth_routes.py` / `routes.py`, and the templates / JS / tests.
- Failure-path smoke (5 scripted reproductions) on the worktree.

## What works well — kept these in mind when grading the rest

- **Additive design is the right call.** Local auth, API keys and the legacy bearer token remain untouched. The `source` kwarg on `AuthManager.create_session` is metadata-only — no decorator behaviour changes — which is exactly the surface where mistakes would be expensive.
- **`oidc` in `SETTINGS_REJECT_KEYS`** is the right place to drop it. `test_oidc_block_rejected_by_validate_settings_post` pins the contract.
- **`_strip_masked_values` reuse for the round-trip** means the GET→edit→POST cycle doesn't accidentally clobber the on-disk `client_secret`. `test_settings_post_roundtrip_preserves_masked_secret` covers it.
- **`_safe_next` is solid.** Smoke-tested 12 vectors (protocol-absolute, protocol-relative, quad-slash, backslash, leading control char, `javascript:`, empty, None, paths with encoding, two backslash variants) — all blocked. Nothing slipped through.
- **`_validate_config` is rigorous** about issuer URL shape, scope list typing, role-mapping schema, and the http-only-for-localhost carve-out (line 197).
- **Test coverage is well chosen.** Reject-list pinned, masking round-trip pinned, JIT vs link vs subject-lookup precedence pinned, claim shapes (scalar + list) pinned, callback failure → /login redirect with sanitised error code pinned.

## Three reproducible findings — would like these addressed before merge

Each item below comes with a Python reproduction (against this PR head, no extra setup beyond a fresh worktree).

### 1. The audit log silently misses every `client_secret` rotation

`modules/web/oidc_routes.py:189-200` compares `before` and `after` snapshots from `get_admin_config()`. But `get_admin_config()` masks `client_secret` to `'********'`. So when an admin POSTs a real new secret, the post-update snapshot also masks it — the `before == after` check for `'client_secret'` always returns True, and the audit entry never lists `client_secret` in `sensitive_changed`.

Reproduction (full suite-context smoke, output verbatim):

```python
from modules.core.oidc import OIDCManager
# build fresh world, enable OIDC with client_secret='SECRET-A'
before = oidc.get_admin_config()
oidc.update_config({'client_secret': 'SECRET-B-rotated'})
after = oidc.get_admin_config()

raw = settings.load_settings()['oidc']
print(f"on_disk_changed = {raw['client_secret'] == 'SECRET-B-rotated'}")   # True
changed = sorted(k for k in after if before.get(k) != after.get(k))
sensitive = [k for k in changed if 'secret' in k]
print(f"changed_keys (route audit)   = {changed}")    # []
print(f"sensitive_changed (route audit) = {sensitive}") # []
```

Result: secret rotated on disk, audit reports zero sensitive changes. A SIEM keying on `details.sensitive_changed` for secret-rotation tracking is blind to every rotation.

**Suggested fix** (one place to change, in `oidc_routes.py::api_oidc_settings_post`): detect rotation by inspecting the request payload before `_strip_masked_values` runs. The cleanest shape:

```python
# After computing `changed` from the masked snapshots, additionally
# detect a client_secret rotation by looking at what the admin sent.
raw_payload_secret = (payload or {}).get('client_secret')
secret_rotated = (
    isinstance(raw_payload_secret, str)
    and raw_payload_secret
    and raw_payload_secret != SECRET_MASK_SENTINEL
)
if secret_rotated and 'client_secret' not in changed:
    changed.append('client_secret')
    sensitive.append('client_secret')
```

(Import `SECRET_MASK_SENTINEL` from `modules.core.settings` or `modules.core.oidc`.)

**Regression test shape**: extend `test_settings_post_roundtrip_preserves_masked_secret` with a sibling case that POSTs a non-sentinel `client_secret` and asserts the captured audit `details['sensitive_changed']` contains `'client_secret'`.

### 2. Role mapping is case-sensitive — capitalised IdP groups silently get `default_role`

`modules/core/oidc.py:444`: `if cv in values`. `values` is `[str(v) for v in raw_value]` — preserves case. Active Directory groups are uppercase (`Domain Admins`); Authentik typically capitalises (`Admin`); Keycloak realm roles depend on how the realm operator typed them. Admins who configure `claim_value: 'admin'` (the canonical lowercase form) will find that capitalised users silently get the default role.

Reproduction:

```python
oidc._map_role({'groups': ['admin']}, cfg)   # 'admin'  (works as expected)
oidc._map_role({'groups': ['Admin']}, cfg)   # 'viewer' (silently downgraded!)
```

`test_map_role_first_match_wins` only exercises lowercase claim values — the bug is invisible to the current suite.

**Suggested fix** (two-line change in `_map_role`):

```python
mapping_values_lower = {str(m.get('claim_value', '')).lower(): m.get('role')
                        for m in cfg.get('role_mappings', [])}
values_lower = [v.lower() for v in values]
for value in values_lower:
    role = mapping_values_lower.get(value)
    if role in VALID_ROLES:
        return role
return default_role
```

This preserves "first match wins" semantics relative to the *claim* values returned by the IdP. If you'd rather preserve "first match wins relative to the admin's mapping order," iterate the mappings as today but lower-case both sides. Either way: comment that the comparison is case-insensitive, since the current code comment explicitly says "case-sensitive" and admins may rely on that.

**Regression test shape**: a third case in `test_map_role_first_match_wins` that passes `'groups': ['Admin']` against the same `eng-admins` / `eng` config — but with one of the mappings using a deliberately mismatched-case value to lock in the new contract.

### 3. End-session URL omits `id_token_hint` → IdP keeps the SSO session after logout

`modules/core/oidc.py:493-513` (`build_end_session_url`) emits at most `?post_logout_redirect_uri=<...>`. The OIDC spec (RP-Initiated Logout, §3) marks `id_token_hint` as RECOMMENDED, and several major IdPs treat it as effectively required:

- Keycloak: rejects logout requests without `id_token_hint` unless an unattended `client_id` parameter is also supplied and the legacy "Allow logout without id_token_hint" toggle is on (off by default in modern versions).
- Okta: silently no-ops the logout if `id_token_hint` is missing.
- Authentik: redirects to the user-confirmation logout page, breaking the headless-SLO assumption.

Net effect today: CertMate kills the local `certmate_session` cookie (good) but the IdP keeps the SSO session alive. The next `Login with SSO` click silently re-authenticates the just-logged-out user with zero credential prompt — which defeats the operator's mental model of "I logged out, the next login will challenge me."

Reproduction:

```python
# simulate the cached Authlib client with discovery metadata
class FakeClient:
    server_metadata = {'end_session_endpoint': 'https://idp.example.com/oauth2/logout'}
oidc._client = lambda app: FakeClient()
print(oidc.build_end_session_url())
# https://idp.example.com/oauth2/logout?post_logout_redirect_uri=https%3A%2F%2Fcertmate.example.com%2Flogin
# (no id_token_hint, no client_id)
```

**Suggested fix** — two-step, because the id_token isn't currently retained past the callback:

1. In `OIDCManager.handle_callback` (after Authlib returns the token):

   ```python
   if token.get('id_token'):
       flask_session['_oidc_id_token'] = token['id_token']
   ```

2. In `OIDCManager.build_end_session_url`:

   ```python
   params = {}
   id_token = flask_session.get('_oidc_id_token')
   if id_token:
       params['id_token_hint'] = id_token
   if cfg.get('client_id'):
       params['client_id'] = cfg['client_id']
   if redirect:
       params['post_logout_redirect_uri'] = redirect
   return f"{end_session}?{urlencode(params)}" if params else end_session
   ```

3. In `auth_routes.py::api_logout`: when invalidating the session, also `flask_session.pop('_oidc_id_token', None)`.

**Regression test shape**: a unit test that stubs `_client` to return a fake metadata object, pre-populates `flask_session['_oidc_id_token']`, calls `build_end_session_url`, and asserts the resulting URL includes `id_token_hint=`. Mirror the structure of `test_settings_post_validates`.

## Two softer observations — flag, not block

### A. PR description overstates the lockout for SSO-only accounts

The PR body says:

> JIT — any IdP-authenticated user gets a `users` row on first login. `password_hash=''` so SSO-only accounts can't fall back to local login.

That's true for the JIT path (`oidc.py:408`) but NOT for the `link_by_email` path (lines 390-400), which preserves the existing `password_hash`. `test_email_collision_links_existing_local_user:262` explicitly asserts the password hash is retained — so the dual-auth behaviour is intentional.

Reproduction:

```python
auth.create_user('alice', 'StrongPw_123!', role='operator', email='alice@example.com')
oidc.resolve_or_provision_user({'iss': '...', 'sub': '...',
                                'email': 'alice@example.com', ...})
# linked into users['alice'], oidc_subject set, password_hash retained
print(auth.authenticate_user('alice', 'StrongPw_123!') is not None)  # True
```

The smooth-migration rationale is fine — users with an existing local password keep it through a rollout. But the description should call out that the lockout applies only to JIT-provisioned identities, since an admin enabling OIDC for MFA-related reasons would want a clear path to "no local fallback for linked accounts" too. Could be either (a) a one-paragraph clarification in the PR body / README, or (b) an explicit admin-controlled toggle `clear_local_password_on_link` (off by default) that admins can flip when they want a hard cutover.

### B. JIT path has an unlocked read-modify-write window

`oidc.py:347-417`: `resolve_or_provision_user` reads `users` via `auth_manager._get_users()`, modifies the local dict, and writes via `auth_manager._save_users()`. No lock. Two concurrent first-time OIDC logins for the *same* candidate username can race:

```
T1 reads users  -> {}
T2 reads users  -> {}
T1 picks 'alice', users['alice'] = {…}
T2 picks 'alice' (no collision in its local view), users['alice'] = {…}
T1 _save_users({alice: T1's record})
T2 _save_users({alice: T2's record})  # last writer wins
```

The blast radius is small (one of the two records is lost, the user retries and the now-existing record is picked up via `_find_by_subject`), but it's the kind of latent surprise that bites a multi-process gunicorn deployment under load. Compare to `update_config` (line 178) which uses `settings_manager.update(mutator)` — that pattern already exists. Routing both linking and JIT through the same atomic mutator closes the window. Not blocking for this PR.

## How I'd like to land this

Happy to take any of the three blockers off your plate via a follow-up PR if a third commit round on this PR is the wrong pace — I've already prototyped the audit-log and role-map fixes locally. Let me know which way you'd prefer.

Bumping to v2.7.0 (not a patch) is right — OIDC is a real surface change. Once these three settle, this is a solid landing.

— Fab

imagreez:
Pushed `5d4b3a3` addressing the three review blockers. Summary of fixes:

### 1. `client_secret` rotation surfaces in the audit log

`modules/web/oidc_routes.py::api_oidc_settings_post`: the masked `before`/`after` snapshot diff was blind to genuine rotations (both come back as `********`). Detection now also inspects the raw payload — when a non-sentinel `client_secret` arrives, `client_secret` is appended to both `changed_keys` and `sensitive_changed`. Regression locked by `test_settings_post_audits_client_secret_rotation`; the masked-sentinel round-trip is pinned NOT to fire a sensitive entry (`test_settings_post_roundtrip_does_not_log_secret_change`).

### 2. Role mapping is case-insensitive

`modules/core/oidc.py::_map_role`: comparison now lowercases both the IdP's claim values and the admin's `claim_value` entries (using a `set` for the claim side). First-match-wins semantics preserved. Regression: `test_map_role_is_case_insensitive` covers lowercase-config↔capitalised-IdP, capitalised-config↔lowercase-IdP, and ordered first-match across cases.

### 3. `id_token_hint` + `client_id` in the end-session URL

- `handle_callback` stashes the raw `id_token` in `flask.session['_oidc_id_token']` (new class constant `_SESSION_ID_TOKEN_KEY`).
- `build_end_session_url` now assembles a parameter dict containing `id_token_hint`, `client_id` and `post_logout_redirect_uri` and only urlencodes the non-empty ones — keeps the bare end_session URL when nothing is available.
- `api_logout` calls a new `clear_session_artifacts()` method after building the URL so a stale id_token can't leak into the next SSO session.

Regression: `test_build_end_session_url_includes_id_token_hint` stubs `_client` with a fake metadata object, populates the flask session, and asserts the URL carries `id_token_hint=…`, `client_id=cm-test` and `post_logout_redirect_uri=`. A sibling test (`test_build_end_session_url_without_id_token`) locks the no-token fallback.

### Soft observation A (PR body lockout claim)

Updated `README.md` "Provisioning and linking" to call out that the no-local-fallback guarantee applies only to JIT-provisioned rows; `link_by_email` preserves the existing password hash so a linked user can still log in locally during rollout. Operators wanting a hard cutover should disable `link_by_email`.

### Soft observation B (JIT race)

Left as-is per your "not blocking" note — happy to route the JIT path through `settings_manager.update(mutator)` in a follow-up PR.

### Test summary

- 5 new tests pinning the three fixes + the round-trip no-op
- Full suite: **897 passed, 10 skipped**, 1 deselected — `tests/test_certificates_concurrency_and_alias.py::TestConcurrentIssuanceLock::test_concurrent_create_attempts_only_one_proceeds`, a barrier-timing test that's already flaky on `928c4c8` (fails ~1 in 5 runs unmodified). Worth a separate ticket; not touched here.

fabriziosalmi:
Yes please — send the follow-up PR for all three blockers (audit-log secret rotation, role-mapping case sensitivity, end-session `id_token_hint`). I'll merge this one once that lands, so v2.7.0 ships with the security surface clean.

Softer items: bundle A (the JIT-vs-link clarification) into the same follow-up if convenient — it's just a README / PR-body tweak. B (JIT race in `resolve_or_provision_user`) I'll open as a separate issue for after v2.7.0; not worth gold-plating this release.

Thanks for the prototyping.

imagreez:
Hey @fabriziosalmi — small clarification: the three blocker fixes are already on this PR's branch, not a follow-up. I think the timing of my push and your reply crossed:

- `5d4b3a3` (`fix(oidc): address PR #206 review blockers`) — pushed to `feature/oidc-sso-191` at 2026-05-19T07:54Z, six minutes before your reply. It shows up in the PR commit list between `928c4c8` and the `main`-merge commit `e24473a` (current head).

So this PR already contains:

- **Blocker 1 — audit on `client_secret` rotation.** `modules/web/oidc_routes.py::api_oidc_settings_post` now inspects the raw payload before strip and appends `client_secret` to both `changed_keys` and `sensitive_changed` when a non-sentinel value arrives. Pinned by `test_settings_post_audits_client_secret_rotation`; the masked round-trip no-op is locked in `test_settings_post_roundtrip_does_not_log_secret_change`.
- **Blocker 2 — case-insensitive role mapping.** `modules/core/oidc.py::_map_role` lowercases both sides (claim values + mapping entries), first-match-wins preserved. Pinned by `test_map_role_is_case_insensitive`.
- **Blocker 3 — `id_token_hint` in the end-session URL.** `handle_callback` stashes the raw id_token in `flask.session['_oidc_id_token']`; `build_end_session_url` emits `id_token_hint` + `client_id` + `post_logout_redirect_uri`; `api_logout` calls a new `clear_session_artifacts()` after building the URL so a stale token can't leak forward. Pinned by `test_build_end_session_url_includes_id_token_hint` and the no-token fallback by `test_build_end_session_url_without_id_token`.
- **Soft A (README).** Added a "Provisioning and linking" note clarifying the no-local-fallback guarantee applies only to JIT-provisioned rows; `link_by_email` preserves the existing `password_hash` and operators wanting a hard cutover should disable it.

Soft B (JIT race in `resolve_or_provision_user`) left as-is for the post-2.7.0 ticket per your note.

OIDC suite on the current PR head: **41/41 green** (`tests/test_oidc_manager.py`, `tests/test_oidc_routes.py`, `tests/test_oidc_settings_masking.py`). Full suite at the time of the push: 897 passed, 1 deselected (pre-existing flaky concurrency test, unrelated).

Status is `mergeable: true / behind` — happy to refresh against the latest `main` (PR #213) whenever you're ready. Let me know if you'd still rather have these as a separate follow-up PR; happy to split if that fits your release flow better, but figured worth checking since the work is already on the branch.

imagreez:
Pushed `493e19f` after a second-pass independent review on the PR head turned up one critical I missed in the first round, plus three smaller items. All fixed in this commit.

## 🔴 Critical: account takeover via `link_by_email` without `email_verified`

`resolve_or_provision_user` step 2 wrote `oidc_subject` onto an existing local user as soon as the IdP returned a matching `email` claim — **with no check on `email_verified`**.

Attack chain (defaults, no config tweaking):
1. Operator enables OIDC against any IdP with self-service signup (public Keycloak realm, Authentik, multi-tenant SaaS where email ownership isn't strictly verified).
2. CertMate has a local `admin` row with `email='admin@target.com'`.
3. Attacker registers an IdP account with email `admin@target.com`, never verifies it.
4. Attacker hits `Login with SSO`. Callback succeeds. `_find_by_email` matches the local admin. `oidc_subject` is written onto the admin row.
5. Next attacker login → authenticates as `admin`. **Role preserved** (linking respects the existing role — that's documented as a feature).

### Fix

- New config field `require_verified_email`, **defaults to True**. Gate fires only at the link seam (no row to take over → no need to gate). Missing claim is treated as not-verified — only an explicit positive attestation links.
- Accepts both bool `True` and JSON string `"true"` (some IdPs serialise this way per OIDC Core §5.1).
- Refused links emit an `oidc_user_link_refused` audit entry with the issuer + email so abuse attempts are surfaced.
- Subject-match path is **not** gated (the IdP's own `sub` is the attestation).
- JIT path is **not** gated (no row to take over).

Operators who trust their IdP to enforce email ownership server-side can flip `require_verified_email=False`.

Regression tests:
- `test_link_refused_when_email_not_verified` — the takeover scenario explicitly.
- `test_link_refused_when_email_verified_claim_missing` — missing != True.
- `test_link_accepted_when_email_verified_is_string_true` — IdP quirk.
- `test_link_refusal_opt_out_via_require_verified_email_false` — escape hatch.
- `test_jit_for_unverified_email_with_no_existing_user` — gate scope.

## Important (non-blocking but addressed)

### JIT race in `resolve_or_provision_user`

You flagged this for follow-up; I'd rather not ship a known data-loss path. The whole read-modify-write now runs inside `settings_manager.update(mutator)`, which already serialises every sequence under the settings `RLock`. The mutator captures its outcome in a closure-bound dict so audit logging (potentially slow I/O) fires **after** `update()` returns rather than under the lock.

Regression test `test_oidc_jit_race_under_concurrent_callbacks`: 8 threads with the same `preferred_username` and 8 distinct subjects, asserts every thread succeeds AND every row lands on disk (no silent overwrites).

### Audit false-positive on `client_secret` resubmit

The detection in `api_oidc_settings_post` flagged `sensitive_changed.client_secret` on ANY non-sentinel value in the payload — even when the admin resubmitted the same plaintext already on disk (paste from a password manager during an unrelated edit, etc.). Pure SIEM noise.

Now reads the unmasked on-disk value before `update_config` runs and only flags a rotation when `raw_payload != prior_plaintext` (in addition to the existing `!= SECRET_MASK_SENTINEL` check). Pinned by `test_settings_post_resubmitting_same_plaintext_does_not_audit_rotation`. The two existing rotation tests still pass — real rotations are still audited.

### `Authlib` pin

`Authlib==1.3.2` → `Authlib>=1.3.2,<2.0.0`. Floor stays at 1.3.2 (GHSA-9c64-2c4r-vf2g was fixed there); ceiling at 2.0.0 because any major bump warrants an explicit compatibility check. Operators pick up future patches and minors without a CertMate bump.

## Test summary

- 48/48 OIDC tests pass (41 prior + 7 new).
- Full suite: 910 passed, 10 skipped. The single failure in the run (`test_concurrent_create_attempts_only_one_proceeds`) is the pre-existing flake already noted in [my earlier comment](https://github.com/fabriziosalmi/certmate/pull/206#issuecomment-4485604246) — passes in isolation.

## Files changed

- `modules/core/oidc.py` — `email_verified` gate, JIT race fix, `_audit` integration.
- `modules/web/oidc_routes.py` — pre-update plaintext capture for rotation detection.
- `requirements.txt` — Authlib pin relaxed.
- `tests/test_oidc_manager.py` — 5 new tests for the verified-email gate + opt-out.
- `tests/test_oidc_routes.py` — 2 new tests (false-positive + JIT race).

Range: [`e24473a..493e19f`](https://github.com/fabriziosalmi/certmate/pull/206/files/e24473a..493e19f).

fabriziosalmi:
Three blockers cleanly resolved — `client_secret` rotation detected from the raw payload so the audit trail is honest, `_map_role` case-insensitive while preserving first-match-wins, and `build_end_session_url` carrying `id_token_hint` + `client_id` plus the cleanup on `/api/auth/logout`. The five regression tests pin each contract.

The standout is in the second commit: catching the `email_verified` gap on the `link_by_email` path is genuinely critical work. That is an account-takeover seam I did not surface in my own playbook review, and you found it doing an independent second-pass on your own PR. The shape of the fix is right — `require_verified_email` defaulted to `True`, the gate scoped to the email-match seam only (subject-match needs no extra gate, JIT has no existing row to take over), refusals emit `oidc_user_link_refused` so an operator can spot abuse attempts. Conservative defaults for the typical org-IdP deployment; operators on a vetted IdP can flip it off knowingly.

JIT race routed through `settings_manager.update(mutator)` is exactly the right shape — reusing the existing RLock rather than introducing a new one, mutator captures outcome in a closure, audit fires after `update()` returns so I/O does not hold the lock. The 8-thread concurrent-callback regression test pins the no-silent-overwrite contract.

The audit false-positive fix on your own previous detection (only flag a real rotation when `raw_payload != prior_plaintext`, not just `!= SECRET_MASK_SENTINEL`) is the right tightening — SIEM noise on a no-op resubmit was the kind of thing that would have eroded operator trust in the signal over time.

CI approved on this revision. Once green I will do a final pass on the new fix surface (email-verified gate, JIT mutator path, audit detection logic) and land for v2.7.0.

Thanks for the rigour — and for the independent second-pass that surfaced the critical you found yourself.

— Fab

fabriziosalmi:
Heads-up: this PR is now `BEHIND` after the v2.7.x prep wave just landed (#214 #215 #216 #217 #219). Required-checks gate is strict-mode, so a rebase against the new main is needed before the merge button accepts it.

Three of the merged PRs touch files this PR also touches:
- `modules/core/settings.py` (your `oidc` block + #215's `_strip_masked_values` empty-string handling + `_deep_merge_dict`)
- `modules/api/resources.py` (your OIDC routes + #219's `validate_domain` in `CreateCertificate.post` and `_validate_domain_path` in `RenewCertificate.post` / `CertificateDNSAliasCheck.get`)

Likely no semantic conflicts — the changes are in non-overlapping sections of each file — but git may still need a manual touch on the merge driver. Once you push the rebase, CI will re-run; I'll watch and land it.

Thanks for the rigour on this round — the email_verified gate, the JIT mutator path, and the audit-noise fix are all the right tightening.

— Fab

fabriziosalmi:
The remaining CodeQL `py/log-injection` thread on `oidc_routes.py:125` is the same false-positive class as every other open log-injection alert on `main` (19 instances of the same shape pre-dating this PR). The value being logged is the IdP-side OAuth error code (e.g. `access_denied`, `interaction_required`) returned in the OIDC callback redirect — a short, controlled-vocabulary string from the IdP, not arbitrary user content. CodeQL flags any expression named `err` that lands in a log line.

The information is also audit-logged via `_audit_failure(..., details={'idp_error': err})` two lines below, so an attacker influencing this value via a crafted IdP callback would surface in the audit trail regardless of what reaches the application log.

Resolving to unblock. The project-wide log-injection class is tracked separately as standing CodeQL noise; the structural fix is to ship a CodeQL config that recognises the project's existing sanitisation patterns, not to per-touch every log line.

fabriziosalmi (approved):
Re-reviewed at commit `e04591d` (post-rebase onto current `main`). Approving for merge.

## What I checked since the prior review

The three blockers from my 2026-05-18 review are all cleanly resolved:

1. **Audit log `client_secret` rotation detection** — fixed in `5d4b3a3`, sanitised against the false-positive resubmit in `1fd3555`. SIEM-facing `sensitive_changed` now fires only on actual rotation.
2. **`_map_role` case-insensitivity** — fixed in `5d4b3a3`. AD `Domain Admins` / Authentik `Admin` / Keycloak realm role capitalisation now matches a single lowercase entry; first-match-wins semantics preserved.
3. **`build_end_session_url` carries `id_token_hint` + `client_id`** — fixed in `5d4b3a3`. Keycloak / Okta / Authentik will actually terminate the IdP-side session on logout. Cleanup on `/api/auth/logout` confirmed.

Plus the standout work from `1fd3555` (independent second-pass review by the author):

4. **`email_verified` gate on `link_by_email`** — closes a CRITICAL account-takeover seam I did not catch. Default `require_verified_email=True`, scoped to the email-match path only, refusals emit `oidc_user_link_refused` audit. Correct shape.
5. **JIT race routed through `settings_manager.update(mutator)`** — uses the existing RLock, audit logs outside the lock. 8-thread regression test pins the no-silent-overwrite contract.
6. **Audit false-positive fix** — only flag a rotation when `raw_payload != prior_plaintext`, not just `!= SECRET_MASK_SENTINEL`. SIEM noise tightened.

## Rebase verification (maintainer push at `e04591d`)

- 12 commits behind `main` resolved cleanly. Single conflict on `modules/__init__.py` (version bump `2.6.11` → `2.7.0`, expected for this PR). All other interactions auto-merged.
- Full local unit suite post-rebase: **957 passed**, 2 skipped, 115 deselected (912 baseline + 45 new OIDC tests).
- CI: `build`, `test (3.12)`, `Analyze python`, `Analyze javascript`, `CodeQL`, `codecov/patch`, `codecov/project` — all SUCCESS.
- Small additional commit `e04591d` (maintainer edit): strips `SETTINGS_REJECT_KEYS` from the test_auth.py round-trip test payload — `oidc` is now reject-listed by design and the prior naive GET → POST shape would 400. The fix mirrors what the real UI does (the OIDC settings have a dedicated mutation endpoint).

## Landing

Squash-merging for v2.7.0. Thanks for the rigour — especially the email_verified finding, which is the kind of self-review I want more of on every release.

PR #217, opened 2026-05-19
feat(dns): issue nested-subdomain wildcards against the parent hosted zone

imagreez:
## Summary

A wildcard cert request for a nested subdomain fails today when the DNS provider only hosts the parent zone. The user-reported scenario:

- Azure DNS hosts `example.com`.
- User asks CertMate for `example2.example.com` + `*.example2.example.com`.
- CertMate strips the wildcard, writes `dns_azure_zone1 = example2.example.com:.../zones/example2.example.com` to `azure.ini`, and `certbot-dns-azure` errors out because no such zone exists in Azure.

The ACME TXT challenge `_acme-challenge.example2.example.com` lives **inside** the `example.com` zone. The plugin can complete the challenge if CertMate passes the actual hosted zone — but `certbot-dns-azure` does not self-discover zones the way Cloudflare, Route53, Google et al. do, so the wiring has to live in CertMate.

This PR adds a per-provider zone-discovery hook so CertMate can ask the provider what zones the account actually hosts, then pick the longest one that covers each cert FQDN — subdomain zone first, parent zone next, as requested.

### New: per-provider zone discovery (`modules/core/dns_zone_discovery.py`)

- `ZoneDiscovery` `Protocol`: `list_zones(account_config) -> list[str]`.
- `AzureZoneDiscovery`: lists Azure DNS zones in the account's resource group via `azure.mgmt.dns.DnsManagementClient` + `azure.identity.ClientSecretCredential`. Both packages are already installed transitively by `certbot-dns-azure` (pinned in `requirements-azure.txt`) — no new top-level dependency.
- `resolve_zones_for_domains(provider, account, fqdns, *, cache=None)`: orchestrates the lookup, dedups longest-first, and raises `ValueError` with an actionable message when an FQDN is not covered or the account has no zones.
- Registry `_ZONE_DISCOVERY` keys discovery hooks by provider. Today only `azure` is registered; adding `rfc2136` or `edgedns` later is a single-line entry. Every other provider falls through to the legacy `_zone_domain` shape (apex-derived from the cert FQDN), preserving their certbot plugin's self-discovery semantics unchanged.

### New helper: `find_covering_zone(fqdn, zones)` (`modules/core/utils.py`)

Pure longest-match-against-a-zone-list helper. Strips a leading `*.` from the FQDN (the TXT challenge for a wildcard lives at the bare apex), normalises trailing dots and case, and applies a **TLD guard** — zones with fewer than two labels (`com`, `tv`, …) are silently skipped. Discovery never surfaces bare TLDs in practice; the guard is defence-in-depth so a hand-edited zone list cannot lead to a whole-TLD match. Multi-label public suffixes (`co.uk`, `com.br`) are not special-cased and match structurally as ≥2-label zones, which is the right behaviour for an operator who genuinely runs a hosted zone there.

### Wired into the cert issuance path (`modules/core/certificates.py`)

- `_dns_config_for_strategy(dns_provider, dns_config, domain, san_domains=None)` now consults the discovery hook for providers in the registry. The matched zones are injected as `_zone_domains: list[str]` (longest-first dedup); per-FQDN matches are logged at `INFO` so the operator can confirm what happened from the run log.
- Both the create path (`create_certificate`) and the renew path (`renew_certificate`) pass SAN lists so a wildcard SAN under a parent zone stays visible at renew time (read from `metadata['san_domains']` on renew).
- Providers absent from the discovery registry keep the legacy `_zone_domain` string injection — zero behavioural change for cloudflare, route53, google, digitalocean, …

### Multi-zone azure.ini (`modules/core/utils.py::create_azure_config`)

Accepts either the legacy single-zone string or a list. When given a list, writes one `dns_azure_zoneN` line per entry; `certbot-dns-azure` performs longest-prefix matching across them at challenge time, picking the most specific zone per FQDN. Empty list raises explicitly rather than writing a zoneless ini that the plugin would later reject with a vaguer error.

### Strategy bridge (`modules/core/dns_strategies.py::AzureStrategy.create_config_file`)

Reads `_zone_domains` (new shape) when present, falls back to `_zone_domain` (legacy shape) otherwise. Every existing test that fed a single string continues to pass without changes; the legacy-keyword regression test (`test_missing_zone_domain_raises_explicit_error`) still locks the actionable error path.

## Test plan

37 new unit tests, all green locally:

- `TestFindCoveringZone` (10) — exact match, subdomain→parent, wildcard prefix stripping, longest-match-wins, case-insensitive, trailing-dot normalisation, TLD guard, label-boundary anchoring (`foo.example.com` does NOT match zone `ample.com`), empty-input safety.
- `TestAzureZoneDiscovery` (4) — Azure SDK mocked at the `DnsManagementClient` boundary: sorted/lowercased return shape, credential argument forwarding, missing-creds `ValueError`, upstream error wrapping with resource-group context, nameless-zone defence.
- `TestResolveZonesForDomains` (6) — single FQDN matches parent zone, SAN list spanning two zones produces both in longest-first order, subdomain-zone wins over parent (user's stated priority), unmatched FQDN raises with both the FQDN and the configured zones in the message, no-zones-at-all path, per-call cache hits exactly once across repeated invocations.
- Registry sanity (2) — `azure` registered, every other provider falls through to the no-op.
- `TestAzureCredentialsFileMultiZone` (3) — multi-zone azure.ini end-to-end through `AzureStrategy.create_config_file`: one `dns_azure_zoneN` per entry, `_zone_domains` takes precedence over `_zone_domain`, empty list raises.
- `TestAzureDnsConfigForStrategy` (6) — the `CertificateManager` bridge: primary wildcard matches parent zone, SAN spanning two zones produces longest-first dedup, subdomain-zone wins, uncovered FQDN raises with actionable message, partial coverage still fails, non-Azure providers bypass discovery entirely.

```bash
python3 -m pytest tests/ -q
# 905 passed, 10 skipped, 0 failed
```

Existing `tests/test_azure_dns_credentials_format.py` (6 tests) keeps passing untouched — the single-zone contract is unchanged on disk.

## Backwards compatibility

- Accounts and certs that already work: no behavioural change. The cert FQDN apex (e.g. `example.com`) appears in the discovered zone list and matches itself, producing the same `dns_azure_zone1 = example.com:...` line as before this PR.
- Accounts that used to fail: the user's reported scenario now succeeds via the parent-zone match.
- Providers other than Azure: zero change. The discovery registry is empty for them, the function returns the legacy `_zone_domain` injection, the strategy reads it via the legacy field.
- A misconfigured Azure SP that lacks `Microsoft.Network/dnsZones/read` on the RG will surface as a runtime error with the resource-group hint baked in — consistent with the fail-early policy.

## Out of scope (follow-ups)

- Discovery hooks for other explicit-zone providers (`rfc2136`, `edgedns`). The registry makes adding them mechanical, but they aren't in the reported scenario.
- Cross-request zone caching. The per-cert-call cache covers SAN batches; if renewal bursts ever become hot, an in-memory TTL cache slots in here.
- Surfacing the resolved zone in the cert detail UI. Logged at `INFO` level today.

### DIFF
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -212,6 +212,15 @@ class CertificateManager:
         against a parent hosted zone — e.g. issuing
         ``*.example2.example.com`` when Azure only hosts ``example.com``.
 
+        **RBAC escape hatch**: if ``dns_config`` carries an explicit
+        ``zone_domains`` list (set by the operator on the account), we
+        skip the live discovery call and use the supplied list directly.
+        That keeps existing Azure service principals working when their
+        scope only includes ``Microsoft.Network/dnsZones/TXT/write`` on
+        specific zones and lacks ``dnsZones/read`` on the resource group
+        — granting the broader read permission for auto-discovery would
+        otherwise be a hard prerequisite for the v2.6.10 upgrade.
+
         For any provider without a discovery hook the legacy single-zone
         shape is preserved: the cert FQDN apex goes into ``_zone_domain``
         and the strategy uses it verbatim. Today that branch is unused
@@ -222,7 +231,10 @@ class CertificateManager:
         if dns_provider != 'azure':
             return dns_config
 
-        from .dns_zone_discovery import has_zone_discovery, resolve_zones_for_domains
+        from .dns_zone_discovery import (
+            has_zone_discovery, resolve_zones_for_domains,
+            resolve_zones_against_explicit_list,
+        )
 
         if has_zone_discovery(dns_provider):
             fqdns = [domain]
@@ -230,20 +242,29 @@ class CertificateManager:
                 for san in san_domains:
                     if san and san not in fqdns:
                         fqdns.append(san)
-            zone_domains = resolve_zones_for_domains(
-                dns_provider, dns_config, fqdns,
+
+            explicit_zones = dns_config.get('zone_domains') if isinstance(dns_config, dict) else None
+            if explicit_zones:
+                # Operator-supplied list — no Azure ARM call needed.
+                # Same matching + fail-early semantics as the discovery
+                # path; just skips the SDK round-trip.
+                zone_domains, per_fqdn = resolve_zones_against_explicit_list(
+                    dns_provider, explicit_zones, fqdns,
+                )
+                source = 'explicit zone_domains'
+            else:
+                zone_domains, per_fqdn = resolve_zones_for_domains(
+                    dns_provider, dns_config, fqdns,
+                )
+                source = 'discovery'
+
+            # One INFO per cert with the full FQDN -> zone map; avoids the
+            # N-line spam a SAN cert with many entries used to produce.
+            logger.info(
+                "Resolved %s DNS zones (%s) for %d FQDN(s): %s",
+                dns_provider, source, len(fqdns),
+                ', '.join(f"{f}->{z}" for f, z in per_fqdn),
             )
-            for fqdn in fqdns:
-                # Mirror the matched zone per FQDN in the log so the
-                # operator can confirm what happened from the run log
-                # without enabling debug. find_covering_zone is cheap.
-                from .utils import find_covering_zone
-                match = find_covering_zone(fqdn, zone_domains)
-                if match:
-                    logger.info(
-                        "Resolved %s DNS zone for %s -> %s",
-                        dns_provider, fqdn, match,
-                    )
             return {**dns_config, '_zone_domains': zone_domains}
 
         # Legacy single-zone fallback (no discovery registered).
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -199,17 +199,54 @@ class CertificateManager:
             return deployment_status
 
     @staticmethod
-    def _dns_config_for_strategy(dns_provider, dns_config, domain):
+    def _dns_config_for_strategy(dns_provider, dns_config, domain, san_domains=None):
         """Return a strategy-ready copy of dns_config with provider-specific extras.
 
-        Azure DNS is the only provider that needs the certificate's primary
-        domain at config-file-creation time (it goes into the
-        ``dns_azure_zone1`` mapping). Injecting it here keeps that knowledge
-        out of the generic flow without forcing every strategy to accept a
-        domain argument.
+        Azure DNS is currently the only provider whose certbot plugin
+        cannot self-discover the hosted zone for an ACME challenge — it
+        wants explicit ``dns_azure_zoneN`` lines in its ini file. We hand
+        it the list of hosted zones the account actually owns (looked up
+        via :func:`modules.core.dns_zone_discovery.resolve_zones_for_domains`)
+        so the plugin's longest-match selects the right zone per
+        challenge. This is what unlocks nested-subdomain wildcards
+        against a parent hosted zone — e.g. issuing
+        ``*.example2.example.com`` when Azure only hosts ``example.com``.
+
+        For any provider without a discovery hook the legacy single-zone
+        shape is preserved: the cert FQDN apex goes into ``_zone_domain``
+        and the strategy uses it verbatim. Today that branch is unused
+        because Azure is the only entry in the registry, but it keeps
+        the contract stable for any future caller / test that still
+        passes the legacy shape.
         """
         if dns_provider != 'azure':
             return dns_config
+
+        from .dns_zone_discovery import has_zone_discovery, resolve_zones_for_domains
+
+        if has_zone_discovery(dns_provider):
+            fqdns = [domain]
+            if san_domains:
+                for san in san_domains:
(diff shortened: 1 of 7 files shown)


imagreez:
Pushed `16cc69c` addressing the self-review findings. Five items + a small test-coverage round.

### 1. Import hoisted out of the per-FQDN loop

`modules/core/certificates.py::_dns_config_for_strategy` — `from .utils import find_covering_zone` no longer runs once per FQDN. Cosmetic but worth not shipping.

### 2. Per-FQDN INFO log collapsed into one line

A SAN cert with 30 entries used to emit 30 `Resolved azure DNS zone for <fqdn> -> <zone>` lines per issuance. Now one INFO carries the full `fqdn->zone` map plus the resolution source (`discovery` vs `explicit zone_domains`). Far more grep-friendly on a busy renewal day.

### 3. RBAC escape hatch — the only item that needed a product call

An Azure service principal whose scope is only `Microsoft.Network/dnsZones/TXT/write` on a specific zone works under the legacy code path but would fail `zones.list_by_resource_group` because that needs `dnsZones/read` on the resource group. Without an escape hatch every upgrade would break those operators on first renewal.

Added an opt-in: if the account carries an explicit `zone_domains` list, `_dns_config_for_strategy` matches against it via the new `resolve_zones_against_explicit_list()` helper and **skips the live Azure ARM call entirely**. Same longest-first matching, same fail-early on a missing match, same INFO log shape.

Operators who hit the regression can recover by adding e.g. `zone_domains: [example.com]` to the affected account in `settings.json`; no RBAC change required. Both Azure-side paths (discovery, explicit) share a single matcher (`_match_fqdns_to_zones`) so their semantics cannot drift.

### 4. Type hint on `create_azure_config`

`zone_domain` was untyped (`Any`). Annotated as `Union[str, List[str]]` to make the dual-shape API explicit.

### 5. Return tuple shape

`resolve_zones_for_domains` and `resolve_zones_against_explicit_list` now return `(matched_zones, per_fqdn_map)` so the caller can render the single consolidated INFO line. Pure refactor — the `matched_zones` ordering is unchanged.

### Test coverage round

* `TestResolveZonesAgainstExplicitList` (5 tests): match-without-discovery (asserts the stubbed discovery hook is NOT invoked), SAN spanning two zones, fail-early on uncovered FQDN, empty-or-whitespace-list rejection, whitespace tolerance on operator-pasted YAML.
* `test_explicit_zone_domains_skip_discovery_call` / `..._still_fail_early_on_missing_match`: bridge integration with the escape hatch active.
* `test_empty_fqdns_raises_explicitly`: defensive — caller bug now surfaces immediately instead of silently writing a zoneless `azure.ini`.
* `test_method_is_only_called_for_dns_challenge_branch`: signature pin so the renewal-path `san_domains` argument cannot be removed without the test failing loudly.

### Acknowledged gaps that I did NOT close (open question for you)

* **HTTP-01 + Azure / `domain_alias` + Azure**: both branches structurally cannot reach `_dns_config_for_strategy`, so they're a negative-space test rather than a positive one. I added the signature pin and documented the invariant in the test file's docstring but did not fake the whole `create_certificate` orbit just to assert non-call. Happy to add full `mock_open`/subprocess fakes if you'd prefer a hard pin — let me know.
* **`renew_certificate` end-to-end with discovery**: the bridge is exercised directly and the `[] or None` collapse on `metadata['san_domains']` is verified by code reading; a full renewal harness would be valuable but is roughly the size of a new fixture and felt out of scope for the review round. Same offer — happy to add if you want it.

### Test summary

* Full suite: **914 passed, 10 skipped, 0 failed**.
* Diff size: +259 / -46 over the original commit (`7af143e`).

Range: [`7af143e..16cc69c`](https://github.com/fabriziosalmi/certmate/pull/217/files/7af143e..16cc69c).

fabriziosalmi:
The 7 unresolved CodeQL `py/incomplete-url-substring-sanitization` threads are all on TEST FIXTURE files (`tests/test_dns_zone_discovery.py`, `tests/test_dns_config_for_strategy_azure.py`) where the test does an `assert <substring> in <url>` to verify the discovered zone string. CodeQL's URL-substring-sanitization rule looks for that string-contains pattern as a sign of broken URL sanitization in production code — it cannot distinguish a test assertion from a runtime defence.

These are not URL sanitizers, they are test assertions. The production code path in `modules/core/dns_zone_discovery.py::find_covering_zone` does proper label-boundary matching (`name == zone or name.endswith('.' + zone)`) and the TLD guard, neither of which is the substring-contains pattern the rule flags. The zone-discovery test contract is correct as written; rephrasing every `assert 'example.com' in result` to silence CodeQL would lose test signal without changing the runtime behaviour.

Resolving these threads to unblock the merge of #217. No change to runtime behaviour or trust boundary; if a real URL-substring-sanitization seam ever appears in production code, CodeQL will surface it on the production file, not on test files.

PR #222, opened 2026-05-19
fix: sanitize certbot stderr before surfacing in API error responses (H3)

fabriziosalmi:
## Summary

Internal security audit (May 2026), finding **H3**: certbot DNS plugins echo the offending `.ini` line verbatim into `stderr` on parse failure. `CertificateManager.create_certificate` and `renew_certificate` wrapped that `stderr` into `RuntimeError(...)`, and the web routes returned `jsonify({'error': str(e)}), 422` — operators reproducing a config bug saw their own credentials echoed in the response body, and any HTTP-traffic logger / SaaS error tracker between them and the API captured the plaintext secret along the way.

## Fix

- New `sanitize_certbot_stderr(stderr_text)` in `modules/core/utils.py`:
  - Regex strips credential-bearing `<name> = <value>` lines and replaces the value with `[REDACTED]`. The name regex anchors on any of `secret|token|password|key|credential|hmac|api_bearer` as a substring of the field name, with **digit-tolerant** surroundings (provider names like `route53` carry digits, so a digit-free character class would have silently skipped `dns_route53_access_key_id` — pinned by a regression test).
  - Credential `.ini` paths replaced with `<credential file>`.
  - Output capped at 4 KB with a truncation marker.
- `CertificateManager.create_certificate` and `renew_certificate` now `logger.error` the raw stderr (operator-visible in application logs) and raise `RuntimeError` with the sanitized copy. Two-leak points, two call sites, same shape.

## Tests

`tests/test_audit_certbot_stderr_sanitizer.py` (new, 9 cases):

- Azure SP `client_secret` line redacted; ACME help URL preserved
- Route53 AWS access key + secret access key both redacted; plugin error narration ("Error 403 from Route53") preserved
- Cloudflare `api_token` redacted
- Generic `smtp_password`, `vault_token`, `api_bearer_token`, `dns_powerdns_api_key` all redacted
- Credential `.ini` paths (relative + absolute) replaced with generic marker
- Non-secret narration (rate limit, ACME URL) survives unredacted
- 10 KB input capped at 4 KB + truncation marker
- `None` / `''` inputs return `''`
- Substring `secret` in non-credential context (e.g. `mysecretrecipe.com` domain) NOT corrupted — pattern requires column-1 `name = value` shape

## Test plan

- [x] `pytest tests/test_audit_certbot_stderr_sanitizer.py` — 9/9
- [x] Full local unit suite: 884 passed, 2 skipped, 115 deselected
- [ ] CI green
- [ ] Manual smoke: misconfigure `dns_azure_sp_client_secret`, trigger create, verify the API 422 body contains `[REDACTED]` not the secret value

### DIFF
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -921,8 +921,16 @@ class CertificateManager:
             )
 
             if result.returncode != 0:
+                # Log the FULL stderr internally for operator debugging.
+                # certbot-dns-azure and a few other plugins echo the
+                # offending credentials .ini line on parse failure, so
+                # the raw stderr carries secret material. Sanitise
+                # before bubbling up into the exception that becomes
+                # the API response body. Internal audit finding H3.
                 logger.error(f"Certbot failed for {domain}: {result.stderr}")
-                raise RuntimeError(f"Certificate creation failed: {result.stderr}")
+                from .utils import sanitize_certbot_stderr
+                safe_stderr = sanitize_certbot_stderr(result.stderr)
+                raise RuntimeError(f"Certificate creation failed: {safe_stderr}")
             
             # Move certificates to standard location
             live_dir = cert_output_dir / 'live' / domain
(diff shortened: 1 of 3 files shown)


fabriziosalmi:
Same false-positive class as #217: the 3 unresolved CodeQL `py/incomplete-url-substring-sanitization` threads are on test fixture lines where the assertion is `assert <substring> in <sanitized_string>` — pinning that operator-facing narration (ACME help URL, plugin error narrative, a domain string) survives the sanitiser unchanged. The pattern triggers the rule even though the assertion is not URL sanitization, it is a test-only contract. Production code in `modules/core/utils.py::sanitize_certbot_stderr` does not use `substring in url` for any decision; the regex anchors on column-1 `name = value` lines exclusively.

Resolving to unblock. No change to runtime behaviour or trust boundary.

Issue #226, opened 2026-05-20
Bug: When no CA chosen and selecting Google Trust Services no options appear. DNS configuration appears in HTTP-01


balkk1:
**Describe the bug**
When within CertMate, so far testing on latest docker pull as of 20th May with a fresh instance. 

<img width="1276" height="549" alt="Image" src="https://github.com/user-attachments/assets/6c62d016-6580-440f-aff6-871d04575d70" />

There isn't any options for settings customisation. But when going back to DNS the block appears for Google.

<img width="1151" height="1256" alt="Image" src="https://github.com/user-attachments/assets/557434ad-46c9-4a46-a3e1-2ef28f8742b9" />

Seems like the drop down box within the CA tab is matching the same within the DNS configuration which causes mismatch.

The lookup within settings.js does a map with google with google-config

Recommended Remediation:
Rename the CA Provider's HTML wrapper element block to something unique, like id="google-ca-config".
Update the lookup dictionary map inside static/js/settings.js to target the updated ID name.

Additional Bug: 
When HTTP-01 is chosen with no CA or settings configured at all DNS configurations gets carried over and no options appear.

<img width="1324" height="955" alt="Image" src="https://github.com/user-attachments/assets/7f6bab27-9084-4a7c-ad92-7f7cdc88862f" />

**To Reproduce**
1. Fresh instance with local account created.
2. Skip Lets encrypt email setup
3. Click on CA tab
4. Choose Google Trust Services
5. You should see no options
6. See error
7. the Display stays stuck on "display:none;"
<img width="1366" height="259" alt="Image" src="https://github.com/user-attachments/assets/4739e0ed-45ea-4b39-9a55-89d536e0fd00" />

Bug 2:
1. Fresh instance with local account created.
2. Skip Lets encrypt email setup
3. Choose any DNS provider
4. Don't setup configuration and click on HTTP-01

**Expected behavior**
I should be able to see the the form to fill information additionally style should dynamically update.

<img width="1385" height="756" alt="Image" src="https://github.com/user-attachments/assets/fb4c6ca6-61f4-4f4e-a0c9-24c79cce2785" />

Bug 2: 
Should be able to see setup for HTTP-01 or options and not see the DNS configuration

**Screenshots**
See above

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Version 148.0.7778.168

PR #227, opened 2026-05-20
fix: Fixes the CA chosen issue where doesn't update accordingly for google since id utilsed elsewhere

balkk1:
Description
Fixed HTML syntax bugs in the Certificate Authority (CA) provider tab component and standardized the element ID conventions for Google Trust Services configuration blocks.

Changes
Fix HTML Syntax: Resolved a broken opening wrapper div (missing >) and removed a stray trailing } character at the base of the file that was causing rendering and template parsing errors.

ID Normalization: Renamed the Google configuration wrapper ID from google-config to google-ca-config to match the project's semantic naming architecture (aligning it with elements like private-ca-config).

Impact
Restores layout styling and resolves broken Alpine.js state reactivity inside the CA Providers view panel.

Eliminates template build compilation issues caused by syntax bugs.
#226

### DIFF
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -217,10 +217,10 @@
             if (!settings.email) {
                 var caDisplayName = defaultCA === 'letsencrypt' ? "Let's Encrypt" :
                     defaultCA === 'zerossl' ? 'ZeroSSL' :
-                        defaultCA === 'google' ? 'Google Trust Services' :
-                            defaultCA === 'buypass' ? 'BuyPass Go' :
-                                defaultCA === 'sslcom' ? 'SSL.com' :
-                                    defaultCA === 'digicert' ? 'DigiCert' : 'Private CA';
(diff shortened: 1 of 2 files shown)


fabriziosalmi:
Thanks for spotting and diagnosing this, @balkk1 — the duplicate `google-config` id was exactly the root cause.

The fix has now landed in `main` via #235, which incorporates your `google-config` → `google-ca-config` rename (credited) and also fixes the second bug from #226: the DNS provider config panels lingering under HTTP-01, because they're siblings of the provider picker rather than children. It adds Playwright regression tests for both.

Note: the HTML-syntax fixes mentioned in this PR's description (the unclosed `<div>` and the stray `}`) don't appear on current `main` — the div is well-formed and there's no trailing brace — so they weren't needed.

Closing in favour of #235 to keep the fix in one place. The credit for the diagnosis is yours — thanks again for the clear write-up!

PR #231, opened 2026-05-20
Fix certificate listing performance and backup pressure

imagreez:
## Summary
- replace per-certificate openssl subprocess parsing with in-process cryptography parsing
- add lightweight cached certificate info retrieval for storage backends, including Azure Key Vault both mode
- reduce routine backup pressure by excluding certbot scratch logs/work while retaining renewal lineage
- document Kubernetes resource guidance for production CertMate deployments

## Tests
- pytest -q
- pytest -q tests/test_metadata_corruption_handling.py tests/test_cert_info_cache_and_storage_summary.py tests/test_azure_keyvault_certificate_storage.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,36 +343,48 @@ For container orchestration and high availability deployments.
 apiVersion: apps/v1
 kind: Deployment
 metadata:
- name: certmate
+  name: certmate
 spec:
- replicas: 2
- selector:
- matchLabels:
- app: certmate
- template:
- metadata:
- labels:
- app: certmate
- spec:
- containers:
- - name: certmate
- image: certmate:latest
- ports:
- - containerPort: 8000
- env:
- - name: API_BEARER_TOKEN
- valueFrom:
- secretKeyRef:
- name: certmate-secrets
- key: api-token
- volumeMounts:
- - name: certificates
- mountPath: /app/certificates
- volumes:
- - name: certificates
- persistentVolumeClaim:
- claimName: certmate-certificates
+  replicas: 1
+  selector:
+    matchLabels:
+      app: certmate
+  template:
+    metadata:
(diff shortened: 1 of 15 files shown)


fabriziosalmi (approved):
Reviewed and tested locally — this is a clean, well-reasoned change. Thanks @rocogamer!

What I checked:

- **In-process parsing**: replacing the per-row `openssl x509 -enddate` subprocess + temp file with `cryptography`'s `x509.load_pem_x509_certificate` is correct, including the timezone handling (`not_valid_after_utc.replace(tzinfo=None)` to match the naive-UTC `utc_now()`). `cryptography` is already a dependency.
- **Cert-info cache**: the key (`domain|renewal_threshold_days|date`) recomputes `days_left`/`needs_renewal` daily and on threshold change, `deepcopy` on get/set avoids aliasing, TTL is env-tunable, and invalidation fires on create/renew/delete/metadata writes. Clearing the whole (small) cache on any metadata mutation is a reasonable, safe choice.
- **Backup**: excluding only the top-level `logs/` and `work/` certbot scratch dirs while keeping `accounts/` and `archive/` preserves renewal lineage across a restore. Correct.
- **Azure both-mode** `retrieve_certificate_info`: avoids the PFX export on the read path (public cert + metadata only) with a careful Secrets/Certificate-API reconciliation and sane fallbacks.
- **resources.py**: switching to `_save_metadata` (cache-invalidating) and returning 500 on failure is an improvement.

Tests: the new/affected suites pass (126 passed, 2 skipped) and the full non-e2e suite is green (969 passed) bar the 3 pre-existing `test_storage_migrate_endpoint.py` failures, which are unrelated and present on `main`. No conflicts with current `main`.

PR #241, opened 2026-05-21
Fix nsone icon during setup wizard

balkk1:


balkk1:
This fixes the NS1 icon during setup wizard. The element named a circle-icon which seem to not exist. Updated to match with RFC icon instead which is suitable.

2026-05-21_20h47_28

image

PR #253, opened 2026-05-24
Added web file routes for /.well-known/acme-challenge/<file from certbot>

rob-infoglobe:
I use certmate behind a reverse proxy server.  I have a backend that routes requests for /.well-known/acme-challenge/* to the certmate docker IP & port.

When using HTTP-01 verification:
  - Certmate's certbot command creates a challenge file in app/data/acme-challenges/.well-known/acme-challenge/
  - Letsencrypt requests /.well-known/acme-challenge/<file> but no route is available to serve the file
  - Letsencrypt gets a 404
  - Challenge doesn't pass, certificate issuance fails

After adding the web route, certmate now serves the static file, challenge/validation passes, & certificate is issued.

### DIFF
--- a/modules/core/factory.py
+++ b/modules/core/factory.py
@@ -341,6 +341,28 @@ def configure_app(container: AppContainer, app, test_config=None):
     app.secret_key = _secret_key_from_env_or_generate(container.data_dir)
     app.config['MAX_CONTENT_LENGTH'] = 50 * 1024 * 1024
 
+    # 1. Define the base ACME storage path
+    app.config['ACME_CHALLENGES_DIR'] = os.environ.get(
+        'ACME_CHALLENGES_DIR', 
+        os.path.join(container.data_dir, 'acme-challenges')
+    )
+
+    # 2. Define the full physical path required for Let's Encrypt
+    # Result: /app/data/acme-challenges/.well-known/acme-challenge/
+    full_acme_path = os.path.join(
+        app.config['ACME_CHALLENGES_DIR'], 
+        '.well-known', 
+        'acme-challenge'
+    )
+
+    # 3. Create the directories if they don't exist
+    try:
+        os.makedirs(full_acme_path, exist_ok=True)
+    except Exception as e:
+        # It's good practice to log this if your logger is available, 
+        # otherwise Flask might fail later when trying to write to this path
+        print(f"Warning: Could not create ACME directory {full_acme_path}: {e}")
+
     if test_config:
         app.config.update(test_config)
 
--- a/modules/web/routes.py
+++ b/modules/web/routes.py
@@ -10,7 +10,7 @@ from functools import wraps
 from pathlib import Path
 from collections import defaultdict
 from time import time
-from flask import request, redirect, url_for
+from flask import request, redirect, url_for, send_from_directory
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


fabriziosalmi (commented):
Thanks for this — it fills a real gap, and the shape is right: a dedicated public route serving the token, using `send_from_directory` (so `werkzeug.safe_join` blocks path traversal) is exactly the safe pattern, and leaving it unauthenticated is correct since the ACME server fetches it anonymously.

There's one thing to resolve before we can merge: the **served path and the path certbot writes to are defined independently**, and they only line up by coincidence in the default Docker layout.

- certbot writes to `HTTP01Strategy.WEBROOT_DIR = 'data/acme-challenges'` (cwd-relative), via `dns_strategies.py` `configure_certbot_arguments` and `certificates.py`.
- this PR serves from `app.config['ACME_CHALLENGES_DIR']`, defaulted to `data_dir/acme-challenges` plus a new `ACME_CHALLENGES_DIR` env.

In Docker (`WORKDIR /app`, data at `/app/data`) both resolve to `/app/data/acme-challenges`, so it works for you. But they diverge if the CWD isn't `/app`, and critically: **setting `ACME_CHALLENGES_DIR` moves the served path but not the path certbot writes to** (the strategy ignores the env), which silently breaks issuance with a 404. The new env knob only wires up half the flow.

### Suggested fix: one resolver, used by all three call sites

Add a single source of truth in `modules/core/dns_strategies.py`:

```python
import os
from pathlib import Path

def acme_webroot_dir() -> Path:
    """Absolute webroot for HTTP-01 challenges. Shared by the certbot
    --webroot arg, the challenge-dir creation, and the serving route, so
    write and serve cannot drift. Override with ACME_CHALLENGES_DIR."""
    return Path(os.environ.get('ACME_CHALLENGES_DIR', 'data/acme-challenges')).resolve()
```

Then have `HTTP01Strategy` use it (drop the `WEBROOT_DIR` class attr):

```python
def configure_certbot_arguments(self, cmd, credentials_file, domain_alias=None):
    cmd.extend(['--webroot', '-w', str(acme_webroot_dir())])
```

`certificates.py` (the challenge-dir pre-creation):

```python
challenge_dir = acme_webroot_dir() / '.well-known' / 'acme-challenge'
challenge_dir.mkdir(parents=True, exist_ok=True)
```

`factory.py` — expose the same resolved path for the route, and log instead of print:

```python
from .dns_strategies import acme_webroot_dir
app.config['ACME_CHALLENGES_DIR'] = str(acme_webroot_dir())
challenge_dir = os.path.join(app.config['ACME_CHALLENGES_DIR'], '.well-known', 'acme-challenge')
try:
    os.makedirs(challenge_dir, exist_ok=True)
except OSError as e:
    logger.warning("Could not create ACME challenge directory %s: %s", challenge_dir, e)
```

The route itself stays as you wrote it (reading `app.config['ACME_CHALLENGES_DIR']`). With this, `ACME_CHALLENGES_DIR` controls both sides and the default behaviour is unchanged.

### Minor
- `print(...)` -> the module `logger` in `factory.py`.
- `routes.py` is missing a trailing newline.

### Tests
Worth a small unit test that the env override propagates to both the certbot `--webroot` arg and the served path (i.e. they stay equal), plus serve-a-token / 404.

I've implemented exactly this on top of your commit (with tests, your commit preserved and co-authored) and it passes locally — happy to push it as a follow-up branch if that's easier for you, or you can apply the above to your fork. Either way, thanks again for the contribution.

rob-infoglobe:
This is great!  Your knowledge of the system is nothing short of impressive.  Thank you for your dedication to this amazing software!  I'm VERY new to github, so I will please defer to your preferred path forward.  It sounds like you've been kind enough to fill the gaps that my fix created.  Thanks again!  :)

fabriziosalmi:
Continuing this in #262, which keeps your commit (co-authored) and adds the one piece we flagged in review: a single `acme_webroot_dir()` resolver shared by certbot's `--webroot`, the challenge-dir creation, and this serving route, so the write and serve paths can't drift (plus the `ACME_CHALLENGES_DIR` override now drives both sides). It ships with tests and passed the full E2E gate including a real issuance + renewal.

Closing this in favour of #262 — thank you for spotting the gap and sending the fix, it's a genuinely useful addition.

PR #270, opened 2026-05-28
perf(certs): reuse loaded settings when listing and scope cert-info cache per domain

imagreez:
## Summary

Two related backend optimizations on the certificate-listing path, plus a
cache-invalidation scoping fix. No API or behavior changes for clients.

### 1. Reuse the already-loaded settings when listing
`CertificateList.get` (`GET /api/certificates`) and the `ZombieScan` loop each
load settings once, then iterate every known domain calling
`certificate_manager.get_certificate_info(domain)` **without** threading that
settings dict through. `get_certificate_info` then re-resolves settings per
domain. The calls are now `get_certificate_info(domain, settings=settings)`,
mirroring the existing `check_renewals` optimization. In-request this avoids a
per-domain deepcopy of the settings dict (`load_settings` is already
request-cached on `flask.g`); in background contexts it avoids real per-domain
disk reads.

### 2. Decouple the cert-info cache from the `settings` argument
`get_certificate_info` previously gated its cross-request
`_certificate_info_cache` on `cache_enabled = (settings is None)`. That meant
optimization #1 would have silently **disabled** the 60s cert-info cache on the
storage-backend path (Azure Key Vault / AWS Secrets Manager / Vault), turning a
cached re-list into N storage round-trips. A new `use_cache` parameter
(default `True`) now controls the cache independently of `settings`, so listing
keeps **both** settings reuse and the cache. `check_renewals` passes
`use_cache=False` to preserve its current behavior (it visits each domain once
per run, so the cache would only add deepcopy-on-set overhead). The
local-filesystem path never used this cache, so it is unaffected.

### 3. Scope cache invalidation per domain
`_invalidate_certificate_info_cache(domain)` called `DeploymentStatusCache.clear()`,
wiping **every** domain's cached info on any single-domain create/renew/metadata
write. It now calls a new `clear_prefix("{domain}|")`, evicting only that
domain's entries. The cache key is always `"{domain}|renewal_threshold_days=…|date=…"`,
so the literal pipe separator keeps prefix matching safe against domains that
share a string prefix (`example.com` vs `example.com.evil`).

## Test plan

Tests run on Linux/Python 3.12 (the suite imports `fcntl`):

- [x] `tests/test_listing_settings_reuse.py` — settings is read once regardless
      of domain count (5 and 40 domains), every per-domain call receives the
      dict, plus a negative-control test proving the un-threaded loop scales
      1-per-domain (guards against silently dropping the kwarg).
- [x] `tests/test_cert_info_cache_and_storage_summary.py` — storage-path cache is
      **retained** when `settings` is threaded (`use_cache` default True: storage
      queried once, second call served from cache) and **bypassed** under
      `use_cache=False`; `clear_prefix` removes only matching keys, isolates the
      pipe-separated prefix, and returns the correct count; a single-domain
      `_invalidate_certificate_info_cache` leaves other domains' entries intact.
- [x] `tests/test_renewal_job_settings_reuse.py` — updated to assert
      `check_renewals` passes `use_cache=False`.
- [x] `pytest` for the above: 16 passed.
- [x] flake8 (`E9,F63,F7,F82`; `--max-complexity=10 --max-line-length=127`): no
      new violations. bandit `--severity-level medium`: no findings.


fabriziosalmi:
Hey @rocogamer — quick heads-up before merge: the CI on this branch surfaced three real regressions, not flakes.

```
tests/test_per_cert_renewal.py::test_check_renewals_skips_disabled_domain
  AssertionError: Only the enabled cert should be renewed; got []

tests/test_per_cert_renewal.py::test_check_renewals_legacy_string_entries_default_to_enabled
  assert [] == ['legacy.example.com']

tests/test_zombie_scanner.py::test_zombie_scan_api_endpoint
  assert 500 == 200
```

Both `per_cert_renewal` failures look like the renewal loop is no longer seeing the `enabled` flag (and legacy string entries) after the settings-reuse path — most likely the reused settings dict is being consumed in a branch that still expects the old read-shape. The zombie scan returning 500 looks like the same shape mismatch surfacing downstream.

Full failed CI: https://github.com/fabriziosalmi/certmate/actions/runs/26564066316

Whenever you have a moment to look would be great. The rest of your queue (#271, #272, #273, #274, #269) is fully green and lined up to merge — this is the only blocker on the certificates.py chain.

Thanks as always for the careful work.

PR #379, opened 2026-07-02
fix(cli): suppress redundant intact reason on audit verify (#378)

### LINKED ISSUE #378: certmate-cli: `audit verify` prints redundant `intact — intact`
### Summary

`certmate audit verify` renders the status word twice when the chain is healthy: **`audit chain: intact — intact`**. The trailing ` — <reason>` is meant to explain a *failure*, but on the happy path the server's `reason` is just `"intact"`, so it duplicates the status. Cosmetic only — the verification itself is correct.

### Steps to reproduce

```console
$ certmate audit verify
audit chain: intact — intact
  signed checkpoint: not cross-checked
```

### Actual vs expected

| | |
|---|---|
| **Actual** | `audit chain: intact — intact` |
| **Expected** | `audit chain: intact` |

(The `signed checkpoint:` line should stay as-is.)

### Root cause

`clients/certmate-cli/certmate_cli/main.py` (`audit_verify`), lines ~288–289:

```python
out.print(f"audit chain: {'[green]intact[/]' if ok else '[red]BROKEN[/]'}"
          f"{(' — ' + reason) if reason else ''}")
```

When `ok` is true the reason (`"intact"`) is redundant with the status label; the ` — reason` suffix is only informative on the `BROKEN` path.

### Suggested fix (~30s)

Only append `reason` when it adds information — e.g. suppress it when it echoes the status:

```python
detail = f" — {reason}" if (reason and reason.lower() != "intact") else ""
out.print(f"audit chain: {'[green]intact[/]' if ok else '[red]BROKEN[/]'}{detail}")
```

(or simply only show the reason on the `not ok` branch).

### Notes
- **Severity: low / cosmetic** — output text only, no behavior change; verification, exit codes, and the `none yet` / `BROKEN` branches are unaffected.
- Shipped in `certmate-cli` 0.1.1. Good candidate to roll into the next patch (0.1.2).

gitcommit90:
## Summary
Fixes #378 — when the audit chain is healthy, `certmate audit verify` no longer prints redundant `intact — intact`; the reason suffix is omitted when `ok` is true and `reason` is `intact` (case-insensitive). BROKEN and `none yet` paths unchanged.

## Test plan
- [x] `pytest tests/test_cli_audit_verify_output.py` (2/2 pass)

### DIFF
--- a/clients/certmate-cli/certmate_cli/main.py
+++ b/clients/certmate-cli/certmate_cli/main.py
@@ -285,8 +285,14 @@ def audit_verify(ctx: typer.Context):
     if not ok and any(k in reason.lower() for k in ("does not exist", "empty")):
         out.print(f"audit chain: [dim]none yet[/] — {reason}")
         return
-    out.print(f"audit chain: {'[green]intact[/]' if ok else '[red]BROKEN[/]'}"
-              f"{(' — ' + reason) if reason else ''}")
+    detail = (
+        f" — {reason}"
+        if reason and not (ok and reason.lower() == "intact")
+        else ""
+    )
+    out.print(
+        f"audit chain: {'[green]intact[/]' if ok else '[red]BROKEN[/]'}{detail}"
+    )
     if cp is not None:
         out.print(f"  signed checkpoint: {'[green]verified[/]' if cp else '[dim]not cross-checked[/]'}"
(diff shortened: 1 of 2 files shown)


fabriziosalmi:
Merged — thanks for fixing your first issue here, and for bringing a regression test with it. Clean work.

PR #392, opened 2026-07-07
release: v2.21.3 (wildcard deployment status + rootless podman)

### LINKED ISSUE #207: Deployment Status Not clear
First off thanks for the last set of update patches and mergers (even if my code wasn't used :))  The fixes seem to work.

This might be a miss-configuration but I do not think so:

Image

The back end is local storage, and yes permissions are 1000:1000 on the entire directory tree.
The certs are downloadable and appear to be correct.

What I noticed is that the deploy script does not run so first thing i did was verify its present and permissions:
```bash
root@docker-01:~# docker exec -ti certmate ls -lR /opt/scripts
/opt/scripts:
total 20
-rwxrwxr-x 1 certmate certmate  790 May 18 18:44 n8n-notify.sh
-rw-rw-r-- 1 certmate certmate 6123 May 12 16:23 reset_password-v2.py
-rw-rw-r-- 1 certmate certmate 6285 May 12 16:08 reset_password.py
```
the n8n-notify.sh is executible, 

```bash
#!/bin/bash

N8N_URL="https://10.10.10.10/webhook/certmate"

# 1. Check for Dry Run
if [ -n "${CERTMATE_DRY_RUN:-}" ]; then
    echo "Dry run detected, skipping webhook notification."
    exit 0
fi

# 2. Concat the certificates to pass to n8n.
CERTMATE_CERT_CONTENT=$(cat $CERTMATE_CERT_PATH)
CERTMATE_KEY_CONTENT=$(cat $CERTMATE_KEY_PATH)
CERTMATE_FULLCHAIN_CONTENT=$(cat $CERTMATE_FULLCHAIN_PATH)

# 3. Execute Webhook
curl -X POST "$N8N_URL" \
     -H 'Content-Type: application/json' \
     -d @- <<EOF
{
    "event": "${CERTMATE_EVENT}",
    "domain": "${CERTMATE_DOMAIN}",
    "cert": "${CERTMATE_CERT_CONTENT}",
    "key": "${CERTMATE_KEY_CONTENT}",
    "fullchain": "${CERTMATE_FULLCHAIN_CONTENT}"
}
EOF
```

When I manually try to `test` the script from the UI it hands back a 127 error

Image

The deploy hooks the most critical feature, as it allows the real deployment of certs in an automated fassion. 

**Enhancement**
the Recent Executions currently have no way of looking into the error, message etc. 
It would be useful if clicking on the status one could see the payload and variables being sent as well as response. This might be helpful in determining if there is an error in the script or in the data being presented. _(this script was tested working in earlier versions)_

### webhook 
while the global hook calling a script works, the labeling could use improvments. 
`Hook name` might be better suited to `path to script or cmd`
and the `e.g. systemctl reload nginx` could probably be better formatted as `Description: e.g. systemctl reload nginx` for better clarity.

I saw this mentioned once before somewhere, but a dedicated `webhook` option might actually be the best solution. Select the `webhook` checkbox and give it a name, enter the URL and method, and optional auth. allow freeform JSON payload and a simple selection of variables that can be populated.

The following screenshot is a snippet from Freshdesk.
selecting advance is just a freeform JSON payload the user creates simple lets you select multiple variables from the drop down.

Image

### LINKED ISSUE #380: add support for podman rootless
Can this be run using podman rootless or podman quadlets?

### LINKED ISSUE #381: Certificates have error displayed under deployment about wrong cert
**Describe the bug**
I raised this issue in #207 but it still exists. When creating a certificate everything works, but the deployment stauts has an icon that indicates that the wrong cert exists. It is unclear why this exists, as the certs appear valid.

**To Reproduce**
Create New cert (server)
provide domain name
enable advance feature and check wildcard
set domain alias
verify alias
create certificate

**Expected behavior**
certificate should create (it does), but it displays error icon see below: This is on every create certificate. 

**Screenshots**

Image

Image

 - OS: MacOS
 - Browser Firefox
 - Version latest

**Smartphone (please complete the following information):**
- NA

**Additional context**
I am guessing this is an error of some sort but it is unclear where that error is so not able to trouble shoot. 

*Unrelated* I stepped away for a few weeks and came back to add a new cert, updated everything to the latest branch and WOW, the whole interface changed. It looks a lot nicer and more polished.  Good work buys. I am shocked at how fast this is evolving. And didn't know where else to give out a huge thank  you!

fabriziosalmi:
## v2.21.3 (Bugfix — wildcard deployment status, rootless-podman / arbitrary-UID support)

Two reported issues from real users.

### Deployment status

- **Wildcard certificates no longer show a false "wrong certificate" error** (#381, recurring from #207). The deployment probe connected to the bare apex for a wildcard — `*.example.com` was probed at `example.com:443` — but a wildcard does not cover its own apex (RFC 6125), so the apex normally serves a different certificate and the fingerprint comparison failed on every wildcard, producing a permanent red "Wrong Cert" that the reporter also could not diagnose. Now: a wildcard without an explicit deployment host is reported as a neutral **"Not Verifiable"** status rather than a mismatch; a new optional per-certificate **`deployment_host`** (settable via `PATCH /api/certificates/<domain>`) lets you point the probe at a covered name (e.g. `www.example.com`) to get a real green check; and every result now carries a diagnostic `mismatch_reason` plus `probe_host`, `served_subject`, `served_fingerprint` and `expected_fingerprint`, wired into the dashboard error-icon tooltip so a mismatch finally explains which host was probed and what did not match. (The domain-alias, suspected in the report, was confirmed to play no role — it only affects DNS-01 validation, not where the certificate is served.)

### Deployment / packaging

- **Rootless podman and arbitrary-UID runtimes are now supported** (#380). The image created its writable directories owner-only (mode 700, UID 1000), so under rootless podman or OpenShift — where the process runs as an arbitrary UID in group 0 — CertMate failed with Permission denied on `/app/data`, `/app/certificates`, and the rest, forcing a manual `chown`. The image now follows the OpenShift arbitrary-UID pattern: the writable trees are group-0 owned and group-writable (setgid), so the app runs correctly as any UID that is a member of group 0, with no host-side chown. The default `USER 1000` is unchanged for normal Docker/Compose, and every secret (CA key, audit signing key, DNS credential files, settings) is still written owner-only 0600. Verified end-to-end running the container as `--user 4711:0` with fresh volumes. New "Rootless podman / OpenShift" section in docs/docker.md covers named-volume (zero-prep), bind-mount, podman-compose and Kubernetes `fsGroup: 0` usage.

## Provenance

Two reported user issues, each fixed at the root and verified:
- **#381** (recurring #207) — wildcard certs showed a permanent false "wrong cert" under Deployment (probe hit the apex, which a wildcard does not cover). Now a neutral "Not Verifiable" status + an optional per-cert `deployment_host` + full diagnostics in the tooltip. The suspected alias was confirmed a red herring.
- **#380** — rootless podman / arbitrary UID: image now follows the OpenShift group-0 pattern (runs as any UID, no manual chown); secrets stay 0600. Verified end-to-end running as `--user 4711:0`.

Local gates: flake8 + bandit + full non-e2e suite green. Mandatory real-cert e2e (LE staging) will run before merge (SENSITIVE_RE matches resources.py + Dockerfile).

Closes #380. Closes #381. Closes #207.


tuxpowered:
Thank you @fabriziosalmi 
Just ran the current version, noticed a few things, one the error message is a bit more informative thank you
image
and 
image

This is on the main server home page. If you click into the item it shows up as the unknown you mentioned.
image

I reissued one of the certs to try and get it to clear with no luck. So i looked into the way the cert was generated: _This is actually how all of my certs are generated_
image

I am guessing that because the certs are issued to `domain.com` and `*.domain.com` that there is an internal check looking at `domain.com` to match the certificate? In my case, we issue the wildcards for different services, but the root domain is usually not under our control. So the cert on `domain.com` would not match the one in certmate, but would still be valid.  If that is the way the checks are being done, and as a way to handle the wildcard checks, maybe when issuing/editing a certificate we could add a monitor/check url to specify the verification endpoint for the cert and if none provided continue looking at the root?

We at least have a little more information on the errors, but I do not think that this moved this specific issue, unless this is the anticipated output.

Issue #437, opened 2026-07-21
Audit-chain rotation needs its own design (split from #431)


fabriziosalmi:
Split out of #431, which covered both halves. The application log is done (PR below): `CERTMATE_LOG_FILE` is opt-in and, when set, is always a `RotatingFileHandler` — there is no code path that produces an unbounded application log.

The audit chain is the half that cannot be solved the same way, and it is worth stating why before anyone reaches for `RotatingFileHandler` here.

## Why rotation is not a one-liner here

`logs/audit/certificate_audit.log` is append-only and **tamper-evident**: every entry chains to the previous one by hash, and `audit verify` recomputes the chain from the beginning. Handing that file to a rotating handler breaks the property the file exists for:

- `RotatingFileHandler` renames the active file and starts an empty one. The new file's first entry has no predecessor, so verification either fails at the seam or has to trust an unverifiable starting point.
- Verification currently reads one file. After a roll, the history spans several, and "the chain is intact" has to mean something across all of them.
- The failure mode is the worst kind: routine maintenance makes `audit verify` report tampering. An operator who sees a tamper alert from a log rotation learns to ignore tamper alerts.

## What a design has to answer

1. **The seam.** Presumably: write a signed checkpoint immediately before the roll, and make the first entry of the new file carry the previous file's head hash as its predecessor — so the chain is continuous across files by construction, not by convention.
2. **Verification across files.** `verify_chain()` and the exported bundle need to walk the set in order, and say which file a break is in. The existing truncated-tail recovery has to keep working per file.
3. **What is safe to delete.** Retention on a tamper-evident record is a policy question, not a disk question: a rotated file that has been deleted is indistinguishable from one that was deleted to hide something, unless the checkpoints attest what existed. Possibly: never delete by default, and require an explicit retention setting that records the deletion in the chain itself.
4. **Compaction as an alternative.** Rotation may be the wrong shape entirely. A periodic signed checkpoint plus archival of verified-and-attested segments might serve the same disk goal without a seam in the live chain.
5. **Migration.** Existing instances have a single-file chain. Whatever the layout becomes, it must verify an old chain unchanged.

## Practical urgency

Low. The chain grows one line per audited operation, so a busy instance measures in tens of MB a year, and the application log — the one that actually grows fast — is now bounded. This should be designed properly rather than shipped quickly; getting it wrong turns the compliance story into a liability.

[...]

PR #442, opened 2026-07-21
fix(audit): make an incremental export slice actually self-verifying (#441)

### LINKED ISSUE #441: Audit export with ?from_seq produces a bundle its own verifier rejects
`GET /api/audit/export?from_seq=N` is documented as producing an "incremental, **self-verifying** slice". It does not: the bundle it returns is rejected by the verifier we tell auditors to use.

## Reproduction

```python
a = AuditLogger(log_dir, chain_dir=chain_dir, signer=AuditSigner(data_dir))
for i in range(6):
    a.log_operation('renew', 'certificate', f'd{i}.example.com', 'success')

bundle = a.export_bundle(from_seq=3)      # 3 entries, seq_first=3
audit_verify.verify_bundle(bundle)
```

```
ok:     false
reason: chain invalid: broken link at seq 3: prev_hash does not match the previous record's hash
```

## Cause

`audit_chain.verify_records()` seeds `prev_hash = GENESIS_PREV` and requires the first record's `prev_hash` to be `""`. Any fragment that does not start at genesis fails the very first link check. There is no way to express "this slice starts at seq N, whose predecessor hash is H".

(That the genesis is mandatory is *correct* and worth keeping — it is why head truncation of a full chain is detected today. The missing piece is not permissiveness, it is an explicit, signed anchor.)

## Why it matters beyond the endpoint

The same missing primitive — *verify a chain fragment against a declared, signed starting anchor* — is what any segmentation or compaction design for #437 needs at its seam. Fixing it here is independent of whichever shape #437 takes, and is a prerequisite for all of them.

## Fix

- `verify_records(records, anchor_prev_hash=GENESIS_PREV, anchor_seq=None)`.
- The manifest carries `anchor_prev_hash` / `anchor_seq` for a non-genesis slice, so the anchor is covered by the bundle signature — the instance *attests* where the slice starts.
- An anchored bundle is `format_version: 2`; genesis-anchored (full) bundles stay byte-identical `format_version: 1`, so existing verifiers keep working on them. An old verifier meeting a v2 bundle reports "unsupported format version" rather than a false "chain invalid".
- The verifier must never report an anchored slice as plain "intact": the output states which seq the verification starts from. An anchored slice proves the entries from the anchor forward are authentic and ordered; it does not prove anything about the prefix, and the CLI has to say so.

Found while analysing #437.

fabriziosalmi:
Fixes #441.

`GET /api/audit/export?from_seq=N` is documented as producing an "incremental, **self-verifying** slice". It produced a bundle that `python -m modules.core.audit_verify --bundle` — the verifier we hand to auditors — rejected:

```
ok:     false
reason: chain invalid: broken link at seq 3: prev_hash does not match the previous record's hash
```

`verify_records()` seeds `prev_hash` with the genesis and requires the first record's `prev_hash` to be `""`. Any fragment fails on its very first link, and it fails claiming "a deletion or reorder" about a perfectly intact record — a false tamper verdict, which is the one output an integrity tool must never produce carelessly.

## Approach

The genesis requirement is **not** relaxed. It is load-bearing: it is why removing entries from the *head* of a full chain is detected today. What was missing is a way for a fragment to declare where it legitimately starts.

- `verify_records(records, anchor_prev_hash=GENESIS_PREV, anchor_seq=None)`. Opt-in, never inferred from the records — a caller must state the anchor.
- An anchored slice is `format_version: 2` and carries `anchor_prev_hash` / `anchor_seq` in the manifest. A full export stays byte-identical `format_version: 1`, so verifiers shipped before this change keep accepting every full export; meeting a v2 bundle they report "unsupported bundle format_version 2" instead of a false "chain invalid".
- The anchor is inside the **signed** manifest, so the signature attests it, and it is checked against the entries rather than merely recorded. Rewriting it breaks both the link check and the signature.
- Downgrading a v2 slice to v1 (to pass a fragment off as a full chain) is rejected explicitly, as is a v2 bundle with no usable anchor.
- Neither the result dict nor the CLI is allowed to call a fragment "intact". Both say **partial slice** and name the anchor seq: it proves the entries from the anchor forward are authentic and ordered, and proves nothing about what came before.

## Tests

`tests/test_audit_export_slice_anchor.py`, 20 tests: the bug itself (mid-chain slice, bounded slice, empty slice, `?from_seq=0`), that anchoring did not open a hole (tampered entry, forged anchor, wrong anchor seq, version downgrade, missing anchor, unknown version), that a slice is never reported as complete (result dict + CLI stderr, both directions), the `verify_records` primitive directly including that the genesis default still rejects a beheaded chain, and the HTTP route end to end — since the route is where the broken workflow was documented.

Suite: 1879 passed, 6 skipped. `docs/api.md` documents the partial-slice semantics and the version compatibility.

## Relation to #437

Found while analysing #437. The same missing primitive — *verify a chain fragment against a declared, signed starting anchor* — is what any segmentation or compaction design needs at its seam, so this lands independently of whichever shape #437 takes.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -779,6 +779,18 @@ The verifier checks the chain structure, that the manifest matches the entries,
 the Ed25519 signature, and that the fingerprint matches the (optionally pinned)
 public key.
 
+**Partial slices.** A full export starts at the genesis and is
+`format_version: 1`. A slice that starts mid-chain (`?from_seq=N` past the first
+entry) is `format_version: 2` and additionally carries `anchor_prev_hash` /
+`anchor_seq` in the manifest — the predecessor hash its first entry continues
+from — so the fragment can be verified even though it has no genesis. The anchor
+is inside the signed manifest, so the signature attests it. The verifier reports
+such a bundle as a **partial slice** and names the anchor seq: it proves the
+entries from the anchor forward are authentic and ordered, and proves nothing
+about what came before. Verifiers older than v2.23.0 report `unsupported bundle
+format_version 2` for an anchored slice; full exports remain byte-compatible
+with them.
+
 > **Threat-model honesty.** The chain + signature detect any interior
 > modification, deletion, or reorder, and tie an export to this instance's
 > public key — for anyone who does not hold the signing key. They do **not**
--- a/modules/core/audit_chain.py
+++ b/modules/core/audit_chain.py
@@ -125,10 +125,25 @@ def verify_chain(path) -> Dict[str, Any]:
     return verify_records(records)
 
 
-def verify_records(records) -> Dict[str, Any]:
+def verify_records(records, anchor_prev_hash: str = GENESIS_PREV,
+                   anchor_seq: Optional[int] = None) -> Dict[str, Any]:
     """Verify a list of parsed chain records ``[{seq, entry, prev_hash, hash}, ...]``.
     Returns the same result dict as :func:`verify_chain`. Shared by the file
-    verifier and the export-bundle verifier."""
+    verifier and the export-bundle verifier.
+
+    By default the records must start at the genesis, i.e. the first record's
+    ``prev_hash`` must be ``""``. That default is load-bearing: it is why
+    removing entries from the *head* of a full chain is detected.
+
+    A **fragment** that legitimately starts mid-chain — an incremental export
+    slice (#441), and the seam of any future segmentation — passes the
+    predecessor hash it continues from as *anchor_prev_hash*, optionally with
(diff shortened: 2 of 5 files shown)


fabriziosalmi:
Good catch — fixed in the follow-up commit.

The scenario mattered more than the tidiness: a v2 slice downgraded by deleting only `anchor_prev_hash` kept `anchor_seq`, passed the check, fell through to the link verification and came back "chain invalid: broken link" — a tamper verdict on an intact record, which is precisely the failure mode this PR exists to remove. Now any anchor field on a v1 bundle is rejected by name, and the test is parametrised over dropping neither, one, or the other field.

Issue #431, opened 2026-07-21
No log rotation anywhere: logs and the audit chain grow without bound


fabriziosalmi:
Found in the 2026-07-21 audit.

`grep -r RotatingFileHandler modules/` returns nothing. `logs/certmate.log`, `logs/audit/certificate_audit.log` and the audit chain grow without bound on a mounted volume.

**Scenario.** A long-lived instance fills its volume; the audit chain is the one file that must not be truncated by hand, so the operator has no safe manual remedy.

**Fix.** Use `RotatingFileHandler` for the application log; for the audit chain, add explicit rotation that preserves and re-anchors the chain (checkpoint, roll, continue) rather than plain truncation.

PR #449, opened 2026-07-22
fix(audit): rotate the human-readable audit log (#443)

### LINKED ISSUE #443: Rotate the audit .log file (the plain one, not the chain)
Part of #437 (approach C).

`AuditLogger.__init__` builds a plain `logging.FileHandler` for `logs/audit/certificate_audit.log` (`audit.py:45`). It never rotates. #431's fix bounded the *application* log through `configure_structured_logging`; this handler is constructed independently and was not covered, so this file is still unbounded in production today.

Unlike the hash chain, this file carries no integrity property: it is human-readable text, it is not hashed or chained, and the `logs/` tree is excluded from backups. Rotating it is a `RotatingFileHandler` swap, nothing more.

**What to do**

- Replace the handler with a `RotatingFileHandler`, defaults consistent with the application log (10 MB × 5), overridable.
- Same failure posture as #431: if the file cannot be opened, warn and keep going. A certificate manager must not refuse to boot over a log file.
- Remove `AuditLogger.get_entries_by_resource()` — it reads the entire file and **has no callers**. Leaving a whole-file reader behind a rotating handler is how a "why is history missing" bug gets born.

**Known, accepted consequence**

`get_recent_entries()` tails only the current file, so immediately after a roll the Activity page and the digest see fewer entries until it refills. At 10 MB ≈ 30k entries versus a digest that asks for 500, this is not worth chasing across backups. Document it; do not build a multi-file reader for it.

Not to be confused with the chain (`data/audit/certificate_audit.chain.jsonl`), which #437 deliberately does not rotate.

fabriziosalmi:
Closes #443. First of the three sub-issues from #437 (approach C).

`AuditLogger.__init__` built a plain `logging.FileHandler` for `logs/audit/certificate_audit.log`. It grew without bound. #431's fix bounded the *application* log through `configure_structured_logging`; this handler is constructed independently and was never reached by it — so the file was still unbounded in production **after** the log-rotation fix shipped.

Safe to rotate because it carries no integrity property: human-readable text, not hashed, not chained, and the `logs/` tree is excluded from backups. The verifiable artifact is the hash chain under `data/audit/`, which is emphatically **not** rotated (#437), and a test pins that rolling the `.log` 200 times leaves the chain a single file that still verifies.

## Also in here

- **An unwritable logs directory no longer stops the application.** This raised out of `__init__`, which the factory calls unguarded — a certificate manager refusing to boot over a *log file*. Now it logs the failure, drops the handler, and keeps writing the chain, which is the record that matters. Tested, including that the chain still verifies with no log handler at all.
- `CERTMATE_AUDIT_LOG_MAX_BYTES` / `CERTMATE_AUDIT_LOG_BACKUP_COUNT`, parsed defensively: `''`, `'lots'`, `'-1'`, `'10MB'` all fall back to the default instead of taking startup down. `max_bytes=0` still buys an unbounded file for an operator who ships it elsewhere.
- **Removed `get_entries_by_resource()`** — it read the entire log, had **no callers**, and behind a rotating handler would have silently returned only what survived in the active file. That is how a "why is the history missing" bug gets born. Dropped from the architecture docs in all five languages.

## Known and accepted

`get_recent_entries()` tails only the active file, so the Activity page and the digest show fewer entries immediately after a roll. At 10 MB ≈ 30k entries against a digest that asks for 500, that is not worth building a multi-file reader for. Documented in the README row.

## Tests

`tests/test_audit_log_rotation.py`, 12 tests: the handler type and defaults, that it actually rolls and `backupCount` caps the tail, that the ` - INFO - ` line format `get_recent_entries` parses is unchanged, env overrides and four unusable env values, explicit `max_bytes=0`, that the hash chain is left alone and still verifies, the unwritable-directory posture, and that the dead reader is gone.

Suite: 1893 passed, 6 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1413,6 +1413,8 @@ main "$@"
 | `CERTMATE_LOG_FILE` |         | -              | Also write logs to this path. Off by default: the container logs to stdout, which is what `docker logs` and log shippers expect. Set it (e.g. `/app/logs/certmate.log`) to keep a file on the mounted volume — it is what the web UI's log stream reads |
 | `CERTMATE_LOG_MAX_BYTES` |    | `10485760`     | Rotate the log file at this size (10 MB). File logging is always rotated — there is no way to configure an unbounded one |
 | `CERTMATE_LOG_BACKUP_COUNT` | | `5`            | How many rotated files to keep (~60 MB ceiling with the default size) |
+| `CERTMATE_AUDIT_LOG_MAX_BYTES` | | `10485760`  | Rotate the human-readable audit log (`logs/audit/certificate_audit.log`) at this size. `0` disables rotation. Does **not** apply to the tamper-evident hash chain in `data/audit/`, which is never rotated |
+| `CERTMATE_AUDIT_LOG_BACKUP_COUNT` | | `5`       | How many rotated audit logs to keep. Note the Activity page tails only the active file, so it shows fewer entries immediately after a roll |
 | `CERTMATE_LOG_LEVEL` |        | `INFO`         | DEBUG / INFO / WARNING / ERROR |
 | `CERTMATE_LOG_JSON` |         | `true`         | JSON log lines (set `false` for human-readable) |
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -605,7 +605,6 @@ log_certificate_downloaded() # Log downloads
 log_batch_operation() # Log batch operations
 log_api_request() # Log API requests
 get_recent_entries() # Get latest audit entries
-get_entries_by_resource() # Get entries for a resource
 ```
 
 ---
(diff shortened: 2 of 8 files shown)


fabriziosalmi:
Second one taken: `_int_env` now says non-negative, and notes that 0 is meaningful rather than sloppy — it is how `logging` spells "never roll".

First one declined, deliberately, and I have added a comment saying why. Propagation to the root handlers means audit lines also reach stdout, which is what `docker logs` and log shippers read; that is pre-existing behaviour and load-bearing here. It is precisely what makes the unwritable-directory path survivable: with `propagate = False` an operator whose `logs/` volume is read-only would lose the audit trail entirely instead of still seeing it on stdout. It is also not duplicate logging in the usual sense — one console handler, one file handler, two different sinks — and "the human-readable audit log is disabled" in the error message refers to the file, which is accurate.

PR #450, opened 2026-07-22
perf(audit): stream chain verification instead of reading the whole file (#444)

### LINKED ISSUE #444: Stream audit-chain verification and export instead of reading the whole file
Part of #437 (approach C).

`audit_chain.verify_chain()` does `f.read().splitlines()`, `load_records()` does the same, and `export_bundle()` materialises every record then `jsonify`s the lot. The chain grows slowly (measured: ~600 KB/year on a busy instance), so this is not urgent — but it is the *only* way chain growth actually hurts today, and it is the reason "we should rotate it" keeps coming up.

Fixing this removes size as an argument for rotation.

**What to do**

- Verify line by line, holding only the running `prev_hash` / `seq` / count. The verifier's contract is unchanged, including the tolerance for a truncated final line and the fail-closed behaviours around checkpoints.
- `load_records(from_seq, to_seq)` filters while streaming instead of building the full list first.
- Consider a streamed response for `/api/audit/export`; at minimum stop holding the file twice (records + serialised JSON).
- The standalone verifier must keep working with stdlib only and no CertMate import.

Behaviour must be identical on every existing test — this is a memory-profile change, not a semantic one.

fabriziosalmi:
Closes #444. Second of the three sub-issues from #437 (approach C).

`verify_chain()` did `f.read().splitlines()`, `load_records()` did the same, and the checkpoint cross-check then read the file a **second** time. The chain is append-only and never truncated — the one file here that only ever grows — so a verifier that must first fit the whole history in memory puts a ceiling on how much history an instance can keep. It is also the reason "we should rotate it" keeps coming up: the pain is in how the file is read, not in the file.

Measured on a 5 MB chain:

```
streaming peak =    25,272 bytes   (0.005 × file size)
buffered  peak = 8,367,501 bytes   (1.673 × file size)
```

## Changes

- `_ChainVerifier` holds the running `prev_hash` / `seq` / count and is fed one record at a time. **Both** the streaming file verifier and the in-memory `verify_records` (used for bundles) run on it, so the two paths cannot drift apart — which matters, because a verifier that disagrees with itself is worse than a slow one.
- `iter_records()` is the lazy form; `load_records()` is now `list(iter_records())` and stays for the one caller that genuinely needs the whole slice at once — building an export bundle, where the bundle *is* the entries and the manifest's `head_hash` is only known after the last of them. `?from_seq`/`?to_seq` is how that one is bounded, and since #441 an anchored slice verifies on its own, so exporting a large chain in pieces is a supported workflow rather than a workaround.
- The checkpoint cross-check scans for one seq and stops, so it takes the generator.

## One deliberate behaviour change

When a file contains more than one fault, the reported one is now the **earliest in the file**. The buffered verifier parsed every line before checking any structure, so an unparseable line late in the file masked a tampered entry early in it. Pinned by a test.

## Tests

`tests/test_audit_chain_streaming.py`, 13 tests. The headline one asserts the property as a number: verifying a ~5 MB chain must peak below a tenth of the file size — it fails loudly against the old implementation (1.67×). The rest pin that behaviour did not move: streamed result identical to `verify_records` on the same records, blank lines anywhere, empty file, truncated last line keeping its tolerant wording, an unparseable interior line still localized by position, the `chain_file_missing` flag that `/api/audit/verify` hangs its absent-vs-tampered decision on, an unreadable path *not* being reported as missing, and the cross-check accepting a generator.

Suite: 1906 passed, 6 skipped. No existing test needed changing.

### DIFF
--- a/modules/core/audit.py
+++ b/modules/core/audit.py
@@ -269,7 +269,13 @@ class AuditLogger:
         instance fingerprint, public key, seq range and head hash; the signature
         is over the canonical manifest, which (via head_hash) transitively
         commits to every entry. ``bundle_signature`` is None when no signer is
-        wired (the entries + chain are still verifiable, just not attributed)."""
+        wired (the entries + chain are still verifiable, just not attributed).
+
+        This one legitimately holds the slice in memory: the bundle *is* the
+        entries, and the manifest's head_hash is only known after the last of
+        them. ``from_seq`` / ``to_seq`` are how a caller bounds it — an
+        anchored slice verifies on its own (#441), so exporting a large chain
+        in pieces is a supported workflow rather than a workaround."""
         with self._chain_lock:
             records = audit_chain.load_records(self.audit_chain_file, from_seq, to_seq)
         signed = self._signer is not None and getattr(self._signer, 'available', False)
@@ -369,7 +375,9 @@ class AuditLogger:
                 "(key rotated, or checkpoints from another instance)"
             )
             return
-        records = audit_chain.load_records(self.audit_chain_file)
+        # Streamed: cross_check_checkpoint scans for one seq and stops, so the
+        # cross-check costs one record of memory, not the whole chain (#444).
+        records = audit_chain.iter_records(self.audit_chain_file)
         check = audit_chain.cross_check_checkpoint(records, latest)
         result["checkpoint_verified"] = bool(check.get("ok"))
         result["checkpoint_seq"] = latest.get("seq")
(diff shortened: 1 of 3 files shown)


fabriziosalmi:
First one taken: the cross-check stops at the checkpointed seq, so it now closes the iterable it abandons in a `finally` — CPython would collect it immediately, but that is an implementation detail to lean on, not a contract. Test added.

Second one declined, and the reasoning is now in the docstring. `iter_records` reads a file that **may have been tampered with**; monotonic `seq` is what the verifier *proves*, not something the reader may presume. Breaking the scan at `to_seq` would mean a reordered chain exports as a quietly-shorter slice — a silent wrong answer traded for a linear scan of a file that measurements put in the hundreds of KB per year. There is a test pinning that a record from beyond `to_seq` appearing early does not truncate the scan. If that scan ever becomes a real cost, the answer is an index, not an assumption.

PR #451, opened 2026-07-22
docs: release notes for v2.22.1

fabriziosalmi:
Release notes for the audit-chain fixes merged since v2.22.0: #442 (#441), #449 (#443), #450 (#444).

Notes land before `scripts/release.sh prepare`, which branches off `origin/main` and requires the section to be there — same order as v2.22.0.

Nothing in the release is breaking: existing chains verify unchanged, and a full export bundle is byte-identical to what previous versions produced.


fabriziosalmi:
Both taken. The intro now says three defects plus the startup hardening that came with #443 — which is what the fourth bullet is — and the env var is spelled out in full with the 10 MB × 5 wording used elsewhere in the file.

Issue #445, opened 2026-07-21
Audit-chain retention: pruning as an explicit, audited operation


fabriziosalmi:
Part of #437 (approach C). This is the half of #437 that is a policy question rather than a disk question.

Retention on a tamper-evident record cannot be a background handler quietly deleting files: a rotated file that has been deleted is indistinguishable from one deleted to hide something. The answer is not to prevent deletion — you cannot — but to make it non-deniable.

**What to do**

- No automatic pruning. Ever, by default.
- An explicit operator action prunes a verified prefix, and **the act of pruning is itself an audit entry in the chain that continues**: `operation="archive"`, `resource_type="audit_chain"`, with the seq range, count and head hash of the exported portion, plus the actor. The record of the deletion survives inside the record it pruned.
- Order is mandatory and must be enforced, not merely documented: export the signed bundle → verify it off the box → only then prune. Refuse to prune a range that has not been exported and verified.
- After pruning, the chain continues with a declared anchor rather than a genesis. #441/#442 added exactly this primitive (`verify_records(anchor_prev_hash=..., anchor_seq=...)`, anchored bundles as `format_version: 2`); the pruned chain file needs the on-disk equivalent, and `verify` must report `verified_from_seq` prominently enough that nobody reads a windowed verdict as a verdict on the whole history.
- A `CERTMATE_AUDIT_CHAIN_WARN_BYTES` threshold that **warns**, so an operator hears about growth without a mechanism silently acting on their audit record.
- `docs/compliance.md` gets the procedure and its limits, in the same honest register as the existing caveats.

This is deliberately the smallest thing that answers "what is safe to delete". The fuller form — a signed carry record with a sealed archive, so the live file verifies on its own and old segments can move off-box (option B in #437) — is a superset of it and can be built on top without a format break.

PR #452, opened 2026-07-22
v2.22.1 - ## v2.22.1 (Audit chain: a false tamper verdict, an unrotated file, a verifier that read everything)

fabriziosalmi:
## v2.22.1 (Audit chain: a false tamper verdict, an unrotated file, a verifier that read everything)

Three defects in the audit trail, plus a startup hardening that came with one of them. All of it was found by taking #437 — "rotating the tamper-evident chain needs its own design" — seriously enough to read the code before writing any. The issue turned out to name the wrong file, and the design question turned out to rest on a primitive that was missing and already broken in production. Full analysis on #437; the remaining half, retention, is #445.

Nothing here is breaking. Existing chains verify unchanged, and a full export bundle is byte-identical to what previous versions produced.

### Audit

- **An incremental export slice verifies instead of accusing you** (#441). `GET /api/audit/export?from_seq=N` was documented as producing a "self-verifying" slice and produced a bundle that `python -m modules.core.audit_verify` — the verifier CertMate hands to auditors — rejected with `broken link at seq N ... a deletion or reorder`. Every documented incremental-export workflow was broken, and it failed in the worst available wording: a tamper verdict on a perfectly intact record. The cause was that verification always started from the genesis, which is load-bearing (it is why removing entries from the *head* of a chain is detected) and is kept. Instead a fragment now declares, inside the **signed** manifest, the predecessor hash it continues from: an anchored slice is `format_version: 2` carrying `anchor_prev_hash`/`anchor_seq`, checked against the entries rather than merely recorded. Neither the API result nor the CLI will call such a slice "intact" — both report a partial slice and name the anchor, because it proves the entries from the anchor forward and nothing at all about the prefix. Full exports remain `format_version: 1` and byte-compatible; a verifier that predates this release meeting an anchored slice says "unsupported format version" rather than claiming the chain is broken.
- **The human-readable audit log is rotated** (#443). `logs/audit/certificate_audit.log` was written through a plain `FileHandler` and grew without bound. The v2.22.0 rotation fix bounded the *application* log; this handler is built independently and was never reached by it, so the file was still unbounded after that release shipped. It is now a rotating handler (`CERTMATE_AUDIT_LOG_MAX_BYTES` / `CERTMATE_AUDIT_LOG_BACKUP_COUNT`, 10 MB × 5 by default, `0` to disable). Safe to rotate precisely because it carries no integrity property: it is text, not hashed, not chained, and excluded from backups. The tamper-evident chain under `data/audit/` is emphatically **not** rotated, and a test pins that rolling the log 200 times leaves it a single intact file. The Activity page tails only the active file, so it shows fewer entries immediately after a roll.
- **An unwritable logs directory no longer stops the application** (#443). Building the audit handler raised out of a constructor the app factory calls unguarded, so a certificate manager refused to start over a *log file*. It now logs the failure, drops the handler, and keeps writing the hash chain — the record that actually matters — which continues to verify with no log file at all.
- **Verifying the chain no longer costs as much memory as the chain is big** (#444). `verify_chain()` read the whole file, `load_records()` read it again, and the checkpoint cross-check read it a third time. The chain is append-only and never truncated, so this put a ceiling on how much history an instance could keep — and made rotation look like the answer to a problem that was really about how the file is read. Verification now streams one record at a time: on a 5 MB chain, peak allocation went from 8.4 MB to 25 KB. The file verifier and the bundle verifier run on the same incremental engine, so the two cannot drift apart. One deliberate behaviour change: when a file contains more than one fault, the reported one is now the earliest in the file rather than whichever the parse pass met first.

### Notes for operators

- `GET /api/audit/verify` and the standalone verifier are unchanged in output for an intact chain. If you script against them, the new fields are `anchored` / `anchor_seq` on bundle verification.
- The two new environment variables are documented in the README table. Neither needs setting; the defaults bound the file.


fabriziosalmi:
CI caught a real defect in `release.sh` itself: `prepare` rewrites the `/health` example version in `README.dockerhub.md` but never staged the file, so it stayed dirty locally and the PR went up without it — and `test_version_consistency`, which pins that example to `modules.__version__`, failed.

The local gates cannot catch this, because they all run *before* the version bump. Same shape as the gate problems v2.22.0 fixed: a step that looked like it had run because nothing said otherwise.

Both the missing bump and the one-line fix to `release.sh` are in this PR.

PR #454, opened 2026-07-22
feat(audit): prune an archived prefix of the chain, explicitly and on the record (#445)

### LINKED ISSUE #445: Audit-chain retention: pruning as an explicit, audited operation

fabriziosalmi:
Closes #445. The last of the three sub-issues from #437 (approach C), at the minimal scope: an operator-initiated prune of an already-exported-and-verified prefix, with the prune itself written into the chain.

Retention on a tamper-evident record is a policy question, not a disk question. You cannot make deletion impossible on a file the operator owns; what you *can* do is make it **non-deniable**. So this is not a background handler that quietly rolls files away.

```bash
# with CertMate stopped, after exporting and verifying the prefix off the box
python -m modules.core.audit_prune --bundle archive-0-1199.json --data-dir data/audit        # dry run
python -m modules.core.audit_prune --bundle archive-0-1199.json --data-dir data/audit --yes
```

## It refuses

Nothing is deleted unless the archive bundle verifies, is signed by this instance, starts exactly where the chain currently starts, and agrees with the chain hash-for-hash at every seq it claims. It also refuses to empty the chain: a chain with no live records cannot be verified at all, only asserted.

**Deliberately CLI-only — there is no API endpoint and there should not be one.** An authenticated request that deletes audit history is precisely what someone who has just compromised an administrator account wants, and an operator who can legitimately prune already has shell access; they have to put the archive somewhere off the box anyway.

## What it leaves behind

- **An `archive` entry in the chain that continues**, naming the seq range, the count, the archived head hash and the archive file's SHA-256. The record of the deletion outlives what it deleted.
- **`certificate_audit.anchor.json`**, signed with the instance key, stating where the chain now starts and what it continues from — so a chain that was merely *truncated* cannot be passed off as a pruned one. Forging it requires the signing key.
- **Verification that never says a bare "intact" again.** `/api/audit/verify` reports `anchored` / `anchor_seq`; the standalone verifier prints that the chain was PRUNED, which entries are not in the file, and whether the anchor signature was actually checked (it is not, unless you pass `--pubkey` — and it says so).

## The two ways this could have gone wrong, both tested

- **An interrupted prune.** The anchor is written *before* the chain is replaced, so a crash between the two loses nothing. Verification then reports "an interrupted prune — re-run it or remove the anchor" instead of a broken chain: the safe direction to fail in.
- **Checkpoints crying wolf.** Signed checkpoints below the anchor attest entries that were just archived; cross-checking against one would report routine maintenance as a truncation — *the* failure mode #437 says must never happen, because an operator who sees a tamper alert from maintenance learns to ignore tamper alerts. Only checkpoints at or above the anchor are used, and a prune past all of them says so plainly rather than failing.

Second and subsequent prunes work, and fall out of #441: on an already-pruned chain the next archive is an anchored slice, which verifies as one.

## Tests

`tests/test_audit_prune.py`, 25 tests, organised as the three properties: it refuses (unverifiable, unsigned, foreign instance, mid-chain slice, would-empty, missing chain, dry run, and that a refusal leaves the file byte-identical); it records itself (the `archive` entry, its head hash, the anchor contents); it does not overclaim or cry wolf (anchored verdicts, the stdlib verifier's wording with and without a pinned key, forged anchor, stranger's key, unreadable anchor failing closed, checkpoints, an anchored chain with no signing key, interrupted prune, planted anchor, growth after a prune, a second prune, and that an export from a pruned chain is anchored).

Suite: 1933 passed, 6 skipped. `docs/compliance.md` documents the procedure and its two limits: the anchor describes only the most recent archive (the full history is the sequence of bundles, which is why each digest is in the chain), and the instance key still does not bind the operator.

### DIFF
--- a/docs/compliance.md
+++ b/docs/compliance.md
@@ -104,6 +104,64 @@ operate certificates on a schedule.
   client-supplied; the trustworthy identity is the authenticated API key.
 - **History boundary.** The chain starts when the feature is first enabled;
   earlier `.log` history is not part of the verifiable chain.
+- **A pruned chain proves less, and says so.** If you have run
+  `audit_prune` (below), verification covers the entries from the anchor
+  forward. The archived prefix is attested by the signed anchor and by the
+  exported bundle you hold off the box — not by the chain file. `verify` reports
+  `anchored` and the anchor seq for exactly this reason; treat a bare "intact"
+  claim about a pruned instance as an overstatement.
+
+## Retention: archiving part of the chain
+
+The chain grows by one line per audited operation — of the order of hundreds of
+kilobytes per year on a busy instance — so most operators never need this. It
+exists for the case where a retention policy requires older records to leave the
+machine.
+
+There is no automatic pruning, and there is no API endpoint for it. Deleting
+audit history is exactly what someone who has just compromised an administrator
+account would want to do, and an operator with a legitimate retention obligation
+has shell access anyway — they have to put the archive somewhere. So it is one
+explicit command, run with CertMate stopped:
+
+```bash
+# 1. Export the prefix you intend to archive (admin token).
+curl -H "Authorization: Bearer $TOKEN" \
+     "https://certmate.example.com/api/audit/export?to_seq=1199" > archive-0-1199.json
+
+# 2. Verify it OFF this box, pinning the instance key.
+python -m modules.core.audit_verify --bundle archive-0-1199.json --pubkey instance.pem
+
+# 3. Store it wherever your policy says. Then, with CertMate stopped:
+python -m modules.core.audit_prune --bundle archive-0-1199.json --data-dir data/audit
+python -m modules.core.audit_prune --bundle archive-0-1199.json --data-dir data/audit --yes
+```
+
+The command without `--yes` is a dry run. It refuses, and removes nothing, if
+the bundle does not verify, is unsigned, is not a prefix of *this* chain, or
+disagrees with the chain at any seq — you never delete records whose only
+remaining copy is unverified.
+
+What it then does is the part that matters for compliance:
+
+- **The deletion is recorded in the chain that survives it.** An `archive` entry
+  is appended naming the seq range, the number of entries, their head hash and
+  the SHA-256 of the archive file. A reader of the pruned chain can see that
+  entries were removed, when, and which archive holds them.
+- **The remainder verifies from a signed anchor.** `certificate_audit.anchor.json`
+  states where the chain now starts and what it continues from, signed with the
+  instance key, so a chain that was simply truncated cannot be passed off as a
+  pruned one — the anchor cannot be forged without the key.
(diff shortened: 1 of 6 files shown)


fabriziosalmi:
Both taken, and the second one was the better catch of the two.

**Interrupted second prune.** After a first prune the chain starts at the previous anchor, not at the genesis — so a second prune dying between writing the new anchor and replacing the chain left a file whose only description had just been overwritten. The verifier fell back to the genesis, hit a broken link at the first record, and reported tampering on a chain from which nothing had been removed. That is precisely the failure #437 exists to prevent, reintroduced by the fix for it.

An anchor now records the one it supersedes (`{anchor_seq, prev_hash}`, `null` on a first prune), signed along with everything else so it cannot be forged. The interrupted state verifies from there and reports `intact from seq 6 ... an interrupted prune`. Two tests: the interrupted second prune, and a chain that starts at neither the anchor nor its predecessor, which fails.

**Vanished anchor.** `_verify_anchor_signature` re-reads the file after `verify_chain` already read it, so a file that disappears in between hit `anchor.get(...)` on `None`. It now fails closed with "the chain anchor disappeared while verifying" rather than raising. Tested with a monkeypatched reader that returns the anchor once and `None` after.

PR #484, opened 2026-07-29
fix: acme-dns issuance never worked — drive it with the native hook (#466)

### LINKED ISSUE #466: ACME-DNS fails
It fail when trying to use acme-dns validation.

Error:
Certificate creation failed: usage: certbot [SUBCOMMAND] [options] [-d DOMAIN] [-d DOMAIN] ... Certbot can obtain and install HTTPS/TLS/SSL certificates. By default, it will attempt to use a webserver both for obtaining and installing the certificate. certbot: error: unrecognized arguments: --acme-dns-credentials <credential file>

Based on https://community.letsencrypt.org/t/problem-with-certbot-dns-acmedns-plugin/219854 it appears that the argument should be "--authenticator dns-acmedns --dns-acmedns-credentials"

fabriziosalmi:
Closes #466.

## The bug

Every acme-dns certificate request failed with:

```
certbot: error: unrecognized arguments: --acme-dns-credentials <credential file>
```

## Root cause

The pinned `certbot-acme-dns==0.3.1` **does** register with certbot — `certbot plugins` lists it — but it exposes only:

- `--acme-dns-server`
- `--acme-dns-propagation-seconds`
- `--acme-dns-is-trusted`

It implements **no credentials-file option at all**, so the command line `AcmeDNSStrategy` built could never parse. acme-dns issuance has been broken for as long as this strategy has existed; it is not a recent regression.

Reproduced against this repo's exact pins (certbot 2.10.0, cryptography 46.0.7, pyopenssl 26.0.0, josepy 1.13.0):

```
$ certbot --help acme-dns
acme-dns:
  Obtains certificates using an ACME DNS server.
  --acme-dns-propagation-seconds ACME_DNS_PROPAGATION_SECONDS
  --acme-dns-server ACME_DNS_SERVER
  --acme-dns-is-trusted

$ certbot certonly --authenticator acme-dns --acme-dns-credentials /tmp/x.json -d example.com --dry-run
certbot: error: unrecognized arguments: --acme-dns-credentials /tmp/x.json
```

That plugin is also simply the wrong shape for CertMate: it registers the acme-dns account *itself*, through interactive `zope.component` display prompts, whereas CertMate stores an already-registered account (`api_url` + `username` + `password` + `subdomain`) in settings and runs certbot headless in a subprocess.

## The fix

Publishing an acme-dns TXT record is a single authenticated POST — and CertMate **already** implements it, in `dns_alias_hook._acme_dns_change`. It was just gated behind explicit DNS-alias mode. This PR routes *every* acme-dns issuance and renewal through that hook.

The alias target is inherently the configured subdomain: acme-dns *is* CNAME delegation, so the user's `_acme-challenge.<domain>` CNAME already points at that subdomain whether or not alias mode was requested.

- **`certificates.py`** — new `_acme_dns_native_alias()` routing switch, used on the create path, on renewal, and to skip the certbot-plugin preflight. Renewal routes on the **provider**, not on metadata, so certificates issued before this fix keep renewing with no metadata migration.
- **`dns_strategies.py`** — `AcmeDNSStrategy.configure_certbot_arguments` now raises instead of emitting flags certbot rejects, so a future refactor cannot silently restore them.
- **`requirements.txt`** — drop `certbot-acme-dns` entirely. It never worked, and it dragged `zope.component`/`zope.interface` into the image for nothing.
- **docs (en/it/de/fr/es)** — acme-dns needs no plugin; also documents the required Subdomain field, which the table omitted.

## Tests

New `tests/test_acme_dns_native_hook.py` (6 tests), each of which fails on `main`:

- plain acme-dns issuance uses `--manual-auth-hook` and emits no `--acme-dns-credentials`
- no certbot-plugin preflight runs for acme-dns
- the hook payload carries the account + subdomain and is deleted afterwards
- **renewal of a pre-fix certificate** (no `domain_alias` in metadata) still takes the hook
- `configure_certbot_arguments` raises
- the routing switch diverts acme-dns only, and normalises/rejects blank subdomains

Full suite: **2220 passed**, 8 skipped, 2 xfailed. `flake8` gate clean, `bandit --severity-level medium` clean.

## Note for reviewers

I could not exercise a live acme-dns server end to end. The certbot argument rejection, the plugin's real option set, and the hook routing are all verified; the actual TXT publish path is the pre-existing `_acme_dns_change` code, unchanged by this PR and already covered by `tests/test_domain_alias.py`.

### DIFF
--- a/docs/de/dns-providers.md
+++ b/docs/de/dns-providers.md
@@ -29,7 +29,7 @@ CertMate unterstützt eine breite Palette von DNS-Anbietern für Let's Encrypt D
 | **Infomaniak** | `certbot-dns-infomaniak` | API-Token | Regional |
 | **ArvanCloud** | `certbot-dns-arvancloud` | API-Schlüssel | Regional |
 | **RFC2136** | `certbot-dns-rfc2136` | Nameserver, TSIG-Schlüssel | Standardprotokoll |
-| **ACME-DNS** | `certbot-acme-dns` | API-URL, Benutzername, Passwort | Spezialisiert |
+| **ACME-DNS** | _integriert_ (kein Plugin) | API-URL, Benutzername, Passwort, Subdomain | Spezialisiert |
 | **Hurricane Electric** | `certbot-dns-he-ddns` | Benutzername, Passwort | Kostenloses DNS |
 | **Dynu** | `certbot-dns-dynudns` | API-Token | Dynamisches DNS |
 | **DuckDNS** | `certbot-dns-duckdns` | Konto-Token | Kostenloses DDNS (ohne eigene Domain) |
--- a/docs/dns-providers.md
+++ b/docs/dns-providers.md
@@ -30,7 +30,7 @@ CertMate supports a wide range of DNS providers for Let's Encrypt DNS-01 challen
 | **Infomaniak** | `certbot-dns-infomaniak` | API Token | Regional |
 | **ArvanCloud** | `certbot-dns-arvancloud` | API Key | Regional |
 | **RFC2136** | `certbot-dns-rfc2136` | Nameserver, TSIG Key | Standard Protocol |
-| **ACME-DNS** | `certbot-acme-dns` | API URL, Username, Password | Specialized |
+| **ACME-DNS** | _built-in_ (no plugin) | API URL, Username, Password, Subdomain | Specialized |
 | **Hurricane Electric** | `certbot-dns-he-ddns` | Username, Password | Free DNS |
 | **Dynu** | `certbot-dns-dynudns` | API Token | Dynamic DNS |
 | **DuckDNS** | `certbot-dns-duckdns` | Account Token | Free DDNS (no domain required) |
--- a/docs/es/dns-providers.md
+++ b/docs/es/dns-providers.md
@@ -29,7 +29,7 @@ CertMate soporta una amplia gama de proveedores DNS para los desafíos Let's Enc
 | **Infomaniak** | `certbot-dns-infomaniak` | Token API | Regional |
 | **ArvanCloud** | `certbot-dns-arvancloud` | Clave API | Regional |
 | **RFC2136** | `certbot-dns-rfc2136` | Servidor DNS, Clave TSIG | Protocolo estándar |
-| **ACME-DNS** | `certbot-acme-dns` | URL API, Nombre de usuario, Contraseña | Especializado |
+| **ACME-DNS** | _integrado_ (sin plugin) | URL API, Nombre de usuario, Contraseña, Subdominio | Especializado |
 | **Hurricane Electric** | `certbot-dns-he-ddns` | Nombre de usuario, Contraseña | DNS gratuito |
 | **Dynu** | `certbot-dns-dynudns` | Token API | DNS dinámico |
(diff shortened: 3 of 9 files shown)


fabriziosalmi:
## Verified end to end against a live acme-dns server

The gap noted in the PR description is now closed. Tested against the public `auth.acme-dns.io` instance and Let's Encrypt **staging**, driving the real production path (`CertificateManager.create_certificate` / `renew_certificate` with a real `ShellExecutor` — only the settings/DNS managers are stubbed, as in the unit tests).

### 1. The hook publishes the TXT record

```
$ CERTBOT_VALIDATION=3mOLtz16bCc3UOKYt5HJDSOI3H2dJDA8bQMoJAqfUPU \
    python3 modules/core/dns_alias_hook.py --config <acme-dns config> --action auth
$ dig +short TXT d6f59a68-....auth.acme-dns.io @8.8.8.8
"3mOLtz16bCc3UOKYt5HJDSOI3H2dJDA8bQMoJAqfUPU"
```

### 2. Issuance succeeds

With a `CNAME _acme-challenge.acmedns-test.certmate.org -> d6f59a68-....auth.acme-dns.io`:

```
INFO  Using DNS provider: acme-dns with account: production
INFO  DNS alias 'd6f59a68-...' requested for acmedns-test.certmate.org;
      using acme-dns manual hook to create TXT records on the alias zone.
INFO  Certificate created successfully for acmedns-test.certmate.org in 19.71 seconds

subject= /CN=acmedns-test.certmate.org
issuer=  /C=US/O=Let's Encrypt/CN=(STAGING) Artificial Amaranth YE1
notBefore=Jul 29 15:13:49 2026 GMT
```

Note the caller passed **no** `domain_alias` — this is the plain acme-dns flow from #466, the one that previously died on `unrecognized arguments`.

### 3. Renewal of a pre-fix certificate succeeds

The metadata written by that issuance contains no `domain_alias`, so it is byte-for-byte the shape of a certificate issued before this PR — the case the renewal routing has to catch by provider:

```json
{"domain": "acmedns-test.certmate.org", "dns_provider": "acme-dns",
 "challenge_type": "dns-01", "staging": true}
```

```
INFO  Renewing acmedns-test.certmate.org with the native acme-dns hook.
INFO  Certificate renewed successfully

serial=2CF582772D8ED23EC2B23C5285C2ED0C24FB
notBefore=Jul 29 15:16:48 2026 GMT   (was 15:13:49 — genuinely re-issued)
```

The throwaway acme-dns account and the test CNAME are being torn down.

fabriziosalmi:
Both review findings were real and are fixed in 53b511b.

**Copilot — misleading error on a blank subdomain.** Correct, and my own test asserted a behaviour that did not exist (its comment claimed the caller would report "not configured", when in fact it fell through to `AcmeDNSStrategy` and raised "this is a bug in the caller"). `_acme_dns_native_alias` now raises `ValueError` naming the field to fill in, which is how `create_certificate` reports every other validation problem and what the API layer turns into a 4xx. It fails before certbot is invoked. Covered by `test_acme_dns_without_a_subdomain_reports_a_config_error` (blank/missing/whitespace) and `test_acme_dns_missing_subdomain_surfaces_through_create`.

**CodeQL — log injection.** The new renewal log line interpolated `domain` straight from metadata. Scrubbed CR/LF the same way `modules/web/cert_routes.py` already does.

Suite: 2224 passed, 8 skipped, 2 xfailed. flake8 and `bandit --severity-level medium` clean.

Issue #403, opened 2026-07-19
Track & remediate the 146 Trivy container/dependency CVEs (base-image rebuild + #103)


fabriziosalmi:
From the code-scanning triage (375 alerts → 225 dismissed as false-positive/accept-by-policy, 4 fixed in #402). The remaining **146 open alerts are all Trivy** — CVEs in the container image and pinned dependencies, **not CertMate application code**, and none is reachable from CertMate's runtime code paths. Tracking here rather than dismissing, since they represent real (if non-exploitable-in-context) CVEs with external remediation.

### Buckets

**1. Base-image OS packages (~143)** — Debian `trixie` packages in `python:3.12-slim-trixie`: curl/libcurl, util-linux, glibc, krb5, libssh2, linux-pam, ncurses, the perl family (incl. the 2 `critical` perl CVEs CVE-2026-42496 / CVE-2026-8376), acl, systemd, gzip/bzip2, etc. None is invoked by CertMate.
- **Action:** rebuild on the newest `python:3.12-slim-trixie` digest (bump the `sha256` pins in `Dockerfile` lines 6 and 50) to pull current Debian security updates; re-scan to see how many clear. Recurring — a periodic base-image bump keeps this bucket small.

**2. `cryptography` bundled OpenSSL — blocked by #103 (1)** — GHSA-537c-gmf6-5ccf (HIGH): the vulnerable OpenSSL statically linked in the `cryptography` wheel. The fixed `cryptography >= 48.0.1` requires `pyOpenSSL >= 26.2`, which the pinned certbot 2.10 / acme stack cannot take. Unblocked only by the **certbot 5.x migration (#103)**. Exposure documented in SECURITY.md (the vulnerable `PKCS7_verify()` is unreachable — no PKCS#7/S-MIME use; TLS I/O uses CPython's own ssl).

**3. Base-image system `pip` (2)** — CVE-2026-6357, CVE-2026-3219 in the image's system `pip 25.0.1` at `/usr/local`. Only exploitable when installing an attacker-supplied wheel, which the runtime never does.
- **Action (optional):** upgrade the base-image system pip in the Dockerfile build stage.

### Acceptance
- Base image rebuilt on a current digest; bucket #1 re-scanned and shrunk.
- GHSA-537c-gmf6-5ccf remains linked to #103 and closes with it.
- SECURITY.md note kept accurate.

fabriziosalmi:
[...]

Issue #103, opened 2026-04-20
CI: restore Python 3.14 matrix leg once josepy/certbot stack is upgraded


fabriziosalmi:
### Context

CI previously ran the test matrix against both Python 3.12 (production via Dockerfile) and Python 3.14. The 3.14 leg was **removed** in commit [`c10f87f`](https://github.com/fabriziosalmi/certmate/commit/c10f87f)'s follow-up because the pinned runtime stack is not 3.14-compatible:

- `certbot==2.10.0` pulls `acme==3.3.0` which requires `josepy<2`, so we are stuck on `josepy==1.13.0`.
- `josepy==1.13.0` raises at class definition time under Python 3.14's PEP 649 deferred annotation evaluation:
  ```
  ValueError: Field `alg` in JSONObject `Header` has no type annotation.
  ```
- Dockerfile ships on Python 3.12 so production is unaffected.

Running a CI leg against a stack we don't support would either (a) silently skip the tests that actually hit the import, or (b) fail forever. Neither is useful.

### Why this is not just a CI cleanup

The underlying stack upgrade is the same one gating Path A of the 'Zero-Domain Certificates' analysis (IP certificates / Let's Encrypt `shortlived` profile), which requires `certbot>=5.3`. That upgrade cascades into:

- `certbot==2.10.0` → `5.x`
- `acme` + `josepy` → matching 5.x-compatible versions
- ~23 `certbot-dns-*` plugins pinned to `==2.10.0` matching certbot — each one needs a version that works against certbot 5.x, or a replacement
- Full DNS provider matrix re-test

So this issue and Path A share the same upgrade surface.

### Definition of done

- `.github/workflows/ci.yml` matrix includes a Python 3.14 leg again.
- The 3.14 leg is green end-to-end against the (upgraded) requirements pins.
- `tests/test_edgedns_credentials_format.py::test_certbot_credentials_parser_reads_all_four_keys` runs (not skipped) on both 3.12 and 3.14.
- Dockerfile may or may not switch to 3.14 at the same time — decision taken in this issue.

### Link

- Removal commit / reasoning: see follow-up commit after [`c10f87f`](https://github.com/fabriziosalmi/certmate/commit/c10f87f)
- Path A spike analysis: private notes (summary: certbot 5.3+ required for IP certs, same upgrade blocker)

fabriziosalmi:
Status check (v2.4.0): blocker still real, keeping this open.

[...]

PR #485, opened 2026-07-29
fix(docker): upgrade the runtime stage's bundled pip (#403)

fabriziosalmi:
Partial fix for #403.

## What

Trivy reports `pip` 25.0.1 in the published image — CVE-2026-8643, CVE-2026-6357, CVE-2026-3219 — even though the builder stage already runs `pip install -U pip`.

That upgrade only ever touched the **builder's** interpreter. The runtime stage starts `FROM` the same base image with its own bundled pip under `/usr/local/lib/python3.12/site-packages`, and nothing upgraded it. That stale copy is what ships, and what Trivy scans:

```
usr/local/lib/python3.12/site-packages/pip-25.0.1.dist-info/METADATA
```

## How

One command added to the runtime stage's existing `apt-get` layer, so the image gains no extra layer.

The application runs from `/opt/venv` and never invokes this pip, so the change is invisible at runtime. It exists so we stop shipping a known-vulnerable copy to anyone who execs into the container and runs `pip`.

## Scope

This is the only Trivy finding in #403 that we can actually close. Of the 247 open Trivy alerts, the **4 CRITICALs are all `perl-base` 5.40.1-6 with an empty `Fixed Version`** — no patched package exists upstream yet, so no rebuild moves them. The rest are OS-layer packages inherited from the base image.

Not touched here, deliberately: `cryptography` 46.0.7 (GHSA-537c-gmf6-5ccf, HIGH) stays pinned. 48.0.1 requires pyopenssl>=26.2.0, which removes `OpenSSL.crypto.X509Extension`, which `acme==3.3.0` evaluates at import time — certbot then crashes on startup. Documented in SECURITY.md; it unblocks with the certbot 5.x epic (#103).

## Verification

I have no Docker daemon on this machine, so the image build and the Trivy delta are verified by this PR's own `build` and `security-scan` jobs rather than locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,10 +58,19 @@ WORKDIR /app
 # routinely start with `#!/bin/bash` — without bash the kernel cannot resolve
 # the shebang and the script returns exit 127 (issue #207).
 # apt-get upgrade pulls security patches for glibc, zlib, etc.
+#
+# The pip upgrade is this stage's, not the builder's. The builder already runs
+# `pip install -U pip`, but that only patches the BUILDER's interpreter — the
+# runtime stage starts from the same base image with its own bundled pip under
+# /usr/local/lib/python3.12/site-packages, which nothing touched. That stale
+# copy is what the Trivy scan keeps reporting (CVE-2026-8643, CVE-2026-6357,
+# CVE-2026-3219 against pip 25.0.1); the app itself runs from /opt/venv and
+# never uses it. See issue #403.
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


fabriziosalmi:
Good catch, and it contradicted this file's own policy — the base image is pinned by digest a few lines up precisely so a rebuild of the same commit cannot drift. Pinned via a build ARG: `ARG PIP_VERSION=26.1.2` (the version that clears all three CVEs), overridable at build time and bumped deliberately when the next pip advisory lands.

Issue #474, opened 2026-07-28
Audit sink for SIEM (syslog/CEF/HTTP) + richer lifecycle events


fabriziosalmi:
## Problem
The audit trail is a signed, hash-chained JSON event stream, but the only way out today is the file + the signed export bundle. SOC/SIEM integration expects a push in a standard format. The lifecycle event catalogue is also thin.

## Proposal
1. An **audit sink** that streams each audit entry to an external collector, with formats:
   - syslog (RFC 5424)
   - CEF
   - generic HTTP/JSON
   Reuse the existing credential/secret sanitizer; the sink must be failure-isolated (never blocks a certificate operation), like the audit chain.
2. Broaden lifecycle events: add `certificate_deployed` and `certificate_revoked`, a scheduled **renewal summary digest**, and optional per-domain routing — mostly wiring `notify()` at existing lifecycle points.

## Acceptance
- Audit events reach a syslog/CEF/HTTP endpoint in the selected format, with secrets redacted.
- The new events fire; the digest is produced on schedule.

fabriziosalmi:
Done — merged in #480 (M5).

Issue #468, opened 2026-07-28
Certificate inventory: fingerprint-keyed model & store


fabriziosalmi:
## Problem
CertMate only knows about certificates it issued (stored in `settings.json` + `data/certs/`). There is no place to record certificates observed elsewhere.

## Proposal
A persistent certificate **inventory**, keyed by certificate **SHA-256 fingerprint**:

- `fingerprint` (PK), subject, SAN list, issuer, serial, `not_before`, `not_after`, key algorithm/size, signature algorithm
- `endpoints[]` — every `host:port` where this certificate was observed (one cert serving many hostnames = one record, many endpoints)
- `source` — `issued` | `probed` | `ct-log` | `imported`
- `managed` — whether CertMate issued/manages it
- `first_seen` / `last_seen`

Store in SQLite with a migration; dedup by fingerprint; update `last_seen` and merge endpoints on re-observation.

## Acceptance
- Records are created/updated idempotently keyed by fingerprint.
- One certificate observed on N hosts yields one record with N endpoints.
- `first_seen`/`last_seen` tracked; certificates CertMate issued are flagged `managed=true` and linked to their managed entry.

Depends on the deep TLS probe.

fabriziosalmi:
Done — merged in #478 (M4: Certificate Discovery & Inventory).

Issue #467, opened 2026-07-28
Deep TLS certificate probe (full served-cert metadata)


fabriziosalmi:
## Problem
CertMate's TLS reachability probe (the deployment-status badge) opens `host:443` and reads the served certificate, but only compares a fingerprint. To build an inventory we need the full certificate as structured data.

## Proposal
Add a reusable "inspect the certificate served at host:port" function that returns structured metadata:

- Subject CN + full **SAN list**
- **Serial number** and **SHA-256 fingerprint** (stable identity)
- `notBefore` / `notAfter` (+ derived days-until-expiry)
- Full **issuer DN** (not just CN)
- **Public-key algorithm and size** (RSA/ECDSA/Ed25519, bits/curve)
- **Signature algorithm**
- Served chain / intermediates where available

Implementation notes:
- Switch `ssl.getpeercert()` → `getpeercert(True)` (DER) + `cryptography` x509 parsing (already a dependency).
- Support IPv6 and a configurable port (not hard-coded 443).
- A cleanly reported status for handshake/verify failures (expired / self-signed / hostname mismatch), not just "error".
- An **SSRF guard**: refuse to probe private/loopback/link-local addresses unless explicitly allowed in config.

## Acceptance
- A function/endpoint returns the fields above for a reachable endpoint, with fixtures covering RSA + ECDSA, with/without SAN, self-signed, and expired certs.
- Private-range targets are rejected by default.

Foundation for the certificate inventory (this milestone).

fabriziosalmi:
Foundation issue for the **Certificate discovery & inventory** milestone. Suggested order:
1. #467 deep TLS probe (this)
2. #468 inventory model & store
3. #469 monitored-endpoint source + #470 CT-log source
4. #471 inventory dashboard
5. #472 adopt/manage a discovered cert
6. #473 algorithm inventory & readiness report

Each builds on the ones above.

Issue #465, opened 2026-07-28
PFX generation


mbw2001:
It appears that when I set a PFX password, the cert.pfx is generated (/app/certificates/DOMAIN/cert.pfx) , but it is not part of the downloadable zip file, and it is only generated for server certs, not client certs (where it is most relevant as I see it).

mbw2001:
I see it in the release notes, but I cannot see how to actually download the PFX, it is still not a part of the ZIP, and I don't see where I can download it for the client certificates?

PR #486, opened 2026-07-29
docs: v2.24.1 release notes

### LINKED ISSUE #466: ACME-DNS fails

fabriziosalmi:
Release notes for the upcoming v2.24.1, ahead of `scripts/release.sh prepare 2.24.1` (which reads this file to build the release PR).

Covers what is already merged to main:

- **#484** — ACME-DNS issuance and renewal (closes #466)
- **#485** — the runtime image's bundled pip (part of #403)

## On the framing

The notes state plainly that **ACME-DNS has never worked in any CertMate release**, rather than describing it as a regression. It was an advertised, documented, UI-exposed provider that failed on every request. Someone who tried it months ago and concluded they had misconfigured something deserves to read that they had not.

## Version choice

Patch, not minor, matching the convention the recent releases already set: v2.23.0 was minor because it added a capability (audit prune), v2.22.1 was a patch because it fixed the audit chain. Nothing new ships here — an existing, already-supported provider was made to work, and a dependency was removed. Both commits are typed `fix:`.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,30 @@
+## v2.24.1 (ACME-DNS, which had never worked)
+
+A patch release with one fix that matters to anyone who tried to use ACME-DNS, and one image hygiene change.
+
+The honest framing first: **ACME-DNS issuance has never worked in any CertMate release.** It was listed as a supported provider, documented, and present in the settings UI, and every certificate request through it failed. If you tried it, gave up, and assumed you had misconfigured something, you had not. This release makes it work.
+
+### ACME-DNS issuance and renewal
+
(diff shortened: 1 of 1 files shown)


fabriziosalmi:
Declining this one: bare issue numbers are already the convention in this file, not an inconsistency. The v2.24.0 section directly below references #467, #468, #474 and #465 the same way, and all four are issues rather than PRs — verified via the API. Changing #466 alone to a PR reference would be the thing that breaks consistency. The GitHub UI resolves both forms to the same link either way.

PR #487, opened 2026-07-29
v2.24.1 - ACME-DNS, which had never worked

fabriziosalmi:
## v2.24.1 (ACME-DNS, which had never worked)

A patch release with one fix that matters to anyone who tried to use ACME-DNS, and one image hygiene change.

The honest framing first: **ACME-DNS issuance has never worked in any CertMate release.** It was listed as a supported provider, documented, and present in the settings UI, and every certificate request through it failed. If you tried it, gave up, and assumed you had misconfigured something, you had not. This release makes it work.

### ACME-DNS issuance and renewal

- **ACME-DNS now issues certificates** (#466). Every request previously died with `certbot: error: unrecognized arguments: --acme-dns-credentials`. The cause was the bundled `certbot-acme-dns` plugin: it registers with certbot, which is why the failure reads like a typo, but it implements no credentials-file option at all — it exposes only `--acme-dns-server`, `--acme-dns-propagation-seconds` and `--acme-dns-is-trusted`. The command line CertMate built could therefore never parse. That plugin was also the wrong shape for CertMate: it expects to register the ACME-DNS account itself through interactive prompts, while CertMate stores an account you registered beforehand.

  Publishing an ACME-DNS TXT record is a single authenticated POST, which CertMate already performed in its own DNS hook for alias-mode validation. Both issuance and renewal now go through that hook, and `certbot-acme-dns` is no longer installed in the image at all.

- **Existing ACME-DNS certificates renew with no action from you.** Renewal routes on the provider rather than on stored metadata, so certificates whose metadata predates this release are handled without any migration.

- **A missing Subdomain is now reported as a configuration error.** An ACME-DNS account saved without its subdomain previously produced an internal error naming the wrong culprit. It now names the field to fill in, and fails before certbot is invoked.

### Image

- **The runtime image ships a current pip** (part of #403). The build stage upgraded pip for itself, but the runtime stage carried the base image's bundled pip 25.0.1, which is what vulnerability scanners reported (CVE-2026-8643, CVE-2026-6357, CVE-2026-3219). It is now pinned to 26.1.2 via a `PIP_VERSION` build argument. CertMate runs from its own virtualenv and never used this pip; the change exists so the image stops shipping a known-vulnerable copy.

### Notes for operators

- If you use ACME-DNS, confirm the **Subdomain** field is set on the account. It is the subdomain ACME-DNS returned at registration, and the value your `_acme-challenge` CNAME points at. The provider documentation table previously omitted it.
- Nothing else changes. There is no configuration migration, no new default, and no change to any other DNS provider.


fabriziosalmi:
Fixed in the release branch: all five landing pages now read 2.24.1.

Worth noting this is the second release in a row where that line drifted — #483 fixed the same thing by hand this morning (2.3.0 to 2.24.0). `release.sh` already bumps `README.dockerhub.md` for exactly this reason ("an example that has to be remembered is an example that goes stale"), and the same argument applies here. Follow-up: add the five `docs/*/index.md` version lines to the bump block in `scripts/release.sh`, alongside a `sed -E` portability fix — the PR-title `sed` in that script uses GNU `\?`, which BSD sed on macOS treats as a literal `?`, which is why this PR opened with `v2.24.1 - ## v2.24.1 (...)` as its title.

PR #488, opened 2026-07-29
chore(release): make version drift and the sed bug impossible, not just fixed

fabriziosalmi:
Follow-up to the v2.24.1 release. Two defects surfaced while cutting it, and both had already happened before and been patched by hand.

## What broke

**1. Docs landing pages drift.** The five `docs/**/index.md` pages carry a hand-written "Current Version" line that `scripts/release.sh` did not own. It went stale for v2.24.0 (fixed by hand in #483) and again for v2.24.1 (#487).

**2. The PR-title `sed` is not portable.** It used GNU's `\?`, which BSD sed on macOS treats as a *literal* `?`. No substitution happened, so v2.24.1's release PR opened with the title `v2.24.1 - ## v2.24.1 (ACME-DNS, which had never worked)`.

## Why the script fix alone is not enough

Both bugs are the same failure: a release step that has to be **remembered**. `release.sh` already encodes the counter-argument, in its own comment about `README.dockerhub.md` — *"an example that has to be remembered is an example that goes stale"*. The docs pages were simply never given the same treatment.

So this PR fixes the script **and** adds the gate:

- `release.sh` bumps every `docs/**/index.md`, **globbed** — a translation added later is covered the day it lands, with no edit to the script.
- `sed -E`, which behaves identically on BSD and GNU.
- `test_version_consistency` now pins every docs landing page (also globbed), with a `test_there_are_docs_landing_pages_to_check` guard so an empty glob cannot make the parametrised test vacuously pass — the exact "green no-op gate" this repo's CI hardening set out to eliminate.
- A new test asserts `RELEASE_NOTES.md` contains a section for the current version. `release.sh` refuses to prepare a release without one, but that check only ever runs on the machine cutting the release; this makes the rule hold in CI for any path reaching main.

## Verification

The gate was confirmed to actually fail, not just pass:

```
$ sed -i '' 's/2.24.1/2.24.0/' docs/it/index.md && pytest tests/test_version_consistency.py
AssertionError: docs/it/index.md advertises version '2.24.0' but
modules.__version__ is '2.24.1'. scripts/release.sh bumps this file; if you
are seeing this outside a release, it was edited by hand.
1 failed, 8 passed
```

The title `sed` and the docs-bump regex were both exercised against the real files before committing. `bash -n scripts/release.sh` clean.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -158,6 +158,21 @@ dh.write_text(
     re.sub(r'("version": ")[0-9]+\.[0-9]+\.[0-9]+(")', rf"\g<1>{v}\g<2>", dh.read_text()),
     encoding="utf-8",
 )
+# Same reasoning for the "Current Version" line on every docs landing page, in
+# every language. These were bumped by hand for v2.24.0 and then AGAIN for
+# v2.24.1 (#483, #487) because the script did not own them. Globbed, so a
+# translation added later is picked up without editing this script;
+# test_version_consistency fails the build if one is ever missed.
+docs = sorted(pathlib.Path("docs").glob("index.md")) + sorted(pathlib.Path("docs").glob("*/index.md"))
+for page in docs:
+    text = page.read_text(encoding="utf-8")
+    bumped = re.sub(
+        r"(?m)^(\*\*[^*]+\*\*\s*:?[ \t]*)[0-9]+\.[0-9]+\.[0-9]+[ \t]*$",
+        rf"\g<1>{v}",
+        text,
+    )
+    if bumped != text:
+        page.write_text(bumped, encoding="utf-8")
 PY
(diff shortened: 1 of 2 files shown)


fabriziosalmi:
Both applied.

The release-notes one was a real gap, and a good catch: `notes_section()` matches with `awk -v v="## v$1 "` and `index($0,v)==1`, so the trailing space is part of the contract. A bare `## v2.24.1` heading extracts an empty section and the release dies — while my looser `\b` assertion would have reported green. Now pinned to `^## v{version} \S`, with the reasoning in a comment so nobody "tidies" the space away later.

Verified both halves rather than assuming: stripping the title makes the test fail, and `notes_section` prints 0 lines for the same input.

The glob message now names the real patterns instead of implying a recursive `**`.

PR #493, opened 2026-08-02
fix: Google Cloud DNS-01 never worked — hand certbot the SA JSON (#385)

### LINKED ISSUE #385: Google Cloud DNS-01: credentials file format is likely never accepted by certbot-dns-google
Found during the v2.21.1 provider credential contract-test sweep, left open because the fix changes the credentials format and needs end-to-end verification against a real Google Cloud DNS zone.

**Symptom (expected):** every Google DNS-01 issuance fails loudly at the plugin.

**Root cause:** `certbot-dns-google` 2.10 expects `--dns-google-credentials` to point at the **service-account JSON itself** (it calls `google.auth.load_credentials_from_file` on it). CertMate instead writes an INI containing `dns_google_project_id` / `dns_google_service_account_key` (see `create_google_config` in `modules/core/utils.py`) and passes that INI as the credentials file — a format the plugin never supported.

**Fix direction:** pass the SA JSON path directly as the credentials file (this also simplifies the per-operation SA-key cleanup added in v2.21.1). Needs a real-zone e2e to confirm, since the unit layer mocks the executor and cannot catch this class of bug (same lesson as Porkbun/Vultr/ArvanCloud/Dynu, all fixed in v2.21.1).

**If you have Google DNS-01 working on a current release, please comment** — that would falsify this analysis and we want to know.

fabriziosalmi:
Closes #385.

## The bug

Every Google Cloud DNS-01 issuance failed. CertMate wrote an ini —

```
dns_google_project_id = my-proj
dns_google_service_account_key = /path/to/google-sa-xxxx.json
```

— and passed **that ini** as `--dns-google-credentials`. But `certbot-dns-google` calls `google.auth.load_credentials_from_file` on whatever path it is handed: the credentials file must **be** the service-account JSON. The ini is rejected outright.

The project id is not a field in any file either. The plugin exposes it as its own flag:

```
$ certbot --help dns-google
  --dns-google-credentials DNS_GOOGLE_CREDENTIALS
                        Credentials (ADC). ...
  --dns-google-project DNS_GOOGLE_PROJECT
```

and its own help text calls the credentials argument the *"Path to Google Cloud DNS service account JSON file"*.

## Verification

Against `certbot-dns-google==2.10.0` with this repo's pins (certbot 2.10.0, cryptography 46.0.7, pyopenssl 26.0.0), calling the exact function the plugin calls, with a real RSA-keyed service-account JSON:

```
A) the ini CertMate wrote  -> DefaultCredentialsError: File ... is not a valid json file
B) the SA JSON directly    -> loaded OK, project detected, Credentials returned
```

And the file `create_google_config` produces **after** this change is accepted by that same call, with the project auto-detected from the key.

**What is not verified:** actual TXT record creation against a live Google Cloud DNS zone — that needs a real GCP project and zone, which I do not have. The credential contract is proven; the record-write path is unchanged plugin code. Anyone with a Google Cloud DNS zone confirming an issuance on this branch would close that last gap.

## Why this looks familiar

This is the acme-dns pattern from #466 all over again: an advertised, documented, UI-exposed provider that could never have worked, with users concluding they had misconfigured something. In #385 @RWGHC was walking backwards through releases hunting for a version where Google DNS worked — there is no such version.

## Changes

- `utils.create_google_config` returns the SA JSON path and writes no ini. The hardening from v2.21.1 is untouched: 0600 + `O_EXCL` creation, random per-operation filename, orphan sweep.
- `GoogleStrategy` passes `--dns-google-project` when a project is configured, and omits it when blank (the plugin then reads it from the key).
- `GoogleStrategy.extra_credential_files` is now empty by design — the credentials file *is* the secret, so the existing `finally` already unlinks it. The cleanup tests for create and renew are unchanged and still pass.

## Tests

`test_p2_quick_wins.py` updated to the real contract, plus new coverage: the credentials file must parse as JSON and equal the key verbatim, no ini is written, the project is passed as a flag, and it is omitted when unset. Two tests asserted the old broken shape and would fail on this branch — deliberately rewritten, not deleted.

Full suite: **2232 passed**, 6 skipped. flake8 and `bandit --severity-level medium` clean.

### DIFF
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -1579,9 +1579,11 @@ class CertificateManager:
             raise
         finally:
             domain_lock.release()
-            # Always clean up credential files (even on failure), including
-            # side files the ini references (Google's SA JSON) — the sweep in
+            # Always clean up credential files (even on failure), including any
+            # side files a provider writes alongside the main one — the sweep in
             # create_google_config only mops up crashed runs, not live ones.
+            # (Google needs no side file since #385: its credentials file IS the
+            # service-account JSON, so it is unlinked as the main one.)
             for cred_path in [credentials_file, *extra_credential_files]:
                 if cred_path:
                     try:
--- a/modules/core/dns_strategies.py
+++ b/modules/core/dns_strategies.py
@@ -225,20 +225,35 @@ class AzureStrategy(DNSProviderStrategy):
     # can fall through to it unchanged.
 
 class GoogleStrategy(DNSProviderStrategy):
+    """Google Cloud DNS.
+
+    The credentials file IS the service-account JSON — see create_google_config
+    and issue #385. Nothing else is written, so there are no extra secret files
+    to clean up: the single returned path is the SA key, and the caller's
+    ``finally`` already deletes the credentials file.
+    """
+
     def __init__(self):
-        # Extra secret files created alongside the returned credentials ini.
-        # The create/renew finally blocks read this so the SA JSON (a live
-        # GCP private key) is deleted with the ini instead of sitting on
-        # disk until the orphan sweep.
         self.extra_credential_files: list = []
+        self._project_id: str = ''
 
     def create_config_file(self, config_data: Dict[str, Any]) -> Optional[Path]:
-        config_file, sa_file = create_google_config(
-            config_data.get('project_id', ''),
+        # Stashed for configure_certbot_arguments: the project is a certbot
+        # flag, not a field in any file.
+        self._project_id = str(config_data.get('project_id') or '').strip()
+        return create_google_config(
+            self._project_id,
             config_data.get('service_account_key', ''),
         )
-        self.extra_credential_files = [sa_file]
-        return config_file
+
+    def configure_certbot_arguments(self, cmd: list, credentials_file: Optional[Path],
+                                    domain_alias: Optional[str] = None) -> None:
+        super().configure_certbot_arguments(cmd, credentials_file, domain_alias=domain_alias)
+        # Optional for the plugin, which otherwise reads project_id out of the
+        # SA JSON. Passed when configured so an operator whose key belongs to a
+        # different project than the zone still resolves the right one.
+        if self._project_id:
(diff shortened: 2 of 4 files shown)


fabriziosalmi:
## Verified against a live Google Cloud DNS zone

The gap noted in the PR description is closed. Ran against a real GCP project with a real service account and a real managed zone, driving `certbot-dns-google`'s own `_GoogleClient` with the credentials file **CertMate produces**.

### Before and after, same client, same key

```
A) the ini CertMate used to write  -> PluginError: Error loading credentials file '...'
B) the file this PR produces       -> client built, project = fab-seclog-01
```

### The TXT record is really written and really removed

```
C) add_txt_record(...)             -> OK

$ gcloud dns record-sets list --zone=certmate-385-test
_acme-challenge.gcp-test.certmate.org.  TXT  60  ['"certmate-385-verification-token"']

D) del_txt_record(...)             -> OK
$ gcloud dns record-sets list --zone=certmate-385-test | grep -i txt
(no TXT left)
```

So the full DNS-01 write path works with the credentials shape this PR introduces, and the record is cleaned up afterwards.

**What is still not covered:** the ACME order and validation round trip, which would need the zone delegated to Google's nameservers. That is unmodified plugin and ACME code — the part this PR changes is exactly the credential load and the record write, and both are now proven against the live API rather than a mock.

All test resources (managed zone, service account, IAM binding, local key) were deleted after the run.

fabriziosalmi:
Good catch, and it was a real inconsistency rather than a style point: the docstring justifies the sweep with "anything older than the certbot timeout is dead", but `certificates.py` kills certbot at **1800s** on both the create and renew paths while the call used `_sweep_orphaned_files`' 3600s default. The window was twice the stated rationale, so a live GCP private key from a crashed run sat on disk for an extra half hour.

Now passed explicitly, with the linkage spelled out in a comment so the two numbers cannot drift apart silently again, plus a test pinning it. Sweeping at 1800s cannot race a live issuance — a key older than the timeout belongs to a process that has already been killed.

Issue #491, opened 2026-08-01
Unable to save settings without specifying a notification email address for Let's Encrypt


exterrestris:
**Describe the bug**
Cannot save General, DNS or Storage settings without updating/configuring settings for a CA (Let's Encrypt by default)

**To Reproduce**
1. Go to Settings
2. (Optional) Update any value / Generate a new API Bearer Token
3. Click 'Save Settings'
4. See "Error saving settings: Email address is required in the Let's Encrypt configuration section" toast notification

**Expected behavior**
Setting changes are saved successfully without requiring updates to unrelated settings

**Screenshots**

<img width="2188" height="1333" alt="Image" src="https://github.com/user-attachments/assets/48bd2348-6143-4621-8e1d-453452a38ea4" />

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Firefox
 - Version: 153

**Additional context**
All CA providers seem to have the same behaviour - changing the CA provider then returning to the General tab to save the settings results in a validation error for the newly selected provider

fabriziosalmi:
Fixed, and released in [v2.24.2](https://github.com/fabriziosalmi/certmate/releases/tag/v2.24.2).

Apologies for closing this without a word — the pull request closed it automatically and you deserved a reply first, having taken the time to write it up properly.

Your diagnosis was right, including the part in "Additional context" about it happening for every CA provider after switching. The email address is used to register an ACME account when a certificate is issued; it was never required to *save* settings, and the server never asked for it — `validate_settings_post` does not look at it. The check was purely in the browser, and it gated the whole form on one unrelated field.

It is now enforced only to complete initial setup, matching how the API bearer token check beside it already worked. After that, saving succeeds and you get a warning that issuing with that CA will fail until an email is added. Nothing becomes silently broken: issuance still refuses if the email is genuinely missing when it matters.

```
docker pull fabriziosalmi/certmate:2.24.2
```

Your other two reports are handled as well — #492 is fixed in the same release, and #490 has a question waiting for you about which shape of download you would actually use.

PR #495, opened 2026-08-02
fix: closing the certificate drawer abandons an in-progress edit (#492)

### LINKED ISSUE #492: Cannot easily create a new certificate after aborting edit of an existing certificate due to confusing UX
**Describe the bug**
Attempting to create a new certificate after aborting an edit of an existing certificate is confusing, as at first glance it doesn't seem to be possible without reloading the page

**To Reproduce**
1. Create a certificate and wait for it to be issued
2. Click on the newly created certificate to open the certificate info dialog
3. Click the 'Edit & reissue' action button
4. Close the Edit certificate dialog without changes using the close button or by clicking outside the dialog
5. Click the new certificate button
6. The Edit certificate dialog opens for the certificate that was previously being edited, and cannot be cleared other than via the (small) 'Cancel edit' button in the dialog body

**Expected behavior**
The (empty) create certificate dialog is opened

 - OS: Windwows 11
 - Browser: Firefox
 - Version: 153

**Additional context**
I can see that there is a 'cancel edit' button displayed in the body of the 'Edit certificate' dialog, however the UX of this is questionable for several reasons. Users would typically expect a cancel button to be next to the submit button (i.e. the Reissue Certificate action), and that closing the dialog via any method other than clicking submit abandons any changes. Further, clicking the 'cancel edit' button simply changes the dialog from 'Edit certificate' to 'Create certificate' rather than returning the user to where they were previously were - the dialog still needs to be manually closed.

fabriziosalmi:
Closes #492. Reported by @exterrestris.

## The bug

1. Open a certificate, click **Edit & reissue**
2. Close the drawer with the X, or by clicking outside it
3. Press **New certificate**

The drawer re-opens still bound to the previous certificate — domain field read-only, fields pre-filled — with no obvious way out. As the reporter put it, the only escape was a small "Cancel edit" button inside the form body, which does not even close the drawer.

## Cause

`closeCertDrawer()` hid the drawer and restored focus, but never touched the reissue state, so `reissueEditingDomain` survived the close. The next open inherited it.

## The fix

Closing a dialog is how people abandon an edit, so `closeCertDrawer()` now treats it as one — for every close path, since the X, the scrim and Escape all route through it.

`cancelEditReissue()` gained a not-editing early return. That is what makes the unconditional call safe: closing a half-filled **create** form must not wipe the fields, and only the edit case resets them.

## On the reporter's wider UX point

They also observed that "Cancel edit" sits in the form body rather than next to the submit action, and that using it leaves the drawer open. That is a fair criticism, but it is a layout change rather than a defect, so it is out of scope here — this PR fixes the trap itself: closing the drawer by any normal means now leaves a clean create form behind, which is the stated expected behaviour. Worth a follow-up issue if the button should move or disappear entirely.

## Tests

A source-level guard in `tests/test_frontend_state_regressions.py`, alongside the one added for #491, asserting both halves: `closeCertDrawer` resets the edit mode, and `cancelEditReissue` keeps its not-editing guard.

**Verified it actually fails**: removing the reset call turns it red with

```
AssertionError: closeCertDrawer no longer clears the reissue edit state — that is
#492: the next 'New certificate' re-opens the previous edit
```

Full suite: **2235 passed**, 6 skipped. `theme_codemod --check` clean.

### DIFF
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -2093,6 +2093,10 @@
     }
 
     function cancelEditReissue() {
+        // No-op outside edit mode, so closing the drawer can call this
+        // unconditionally (#492) without wiping a create form the user was
+        // half-way through filling in.
+        if (!reissueEditingDomain) return;
         reissueEditingDomain = null;
         var domainField = document.getElementById('domain');
         domainField.readOnly = false;
--- a/templates/index.html
+++ b/templates/index.html
@@ -439,6 +439,15 @@
             if (focusTarget) setTimeout(function () { focusTarget.focus(); }, 320);
         }
         function closeCertDrawer() {
+            // Leaving the drawer abandons an in-progress "Edit & reissue"
+            // (#492). The edit state used to survive the close, so the next
+            // "New certificate" silently re-opened the edit form for the
+            // previous certificate, with its domain field read-only and no
+            // obvious way back. Closing a dialog is how people abandon an
+            // edit, so treat it as one. No-op when not editing.
+            if (typeof window.cancelEditReissue === 'function') {
+                window.cancelEditReissue();
+            }
             var d = document.getElementById('createCertFormContainer');
             var s = document.getElementById('certDrawerScrim');
             if (d) d.classList.add('translate-x-full');
--- a/tests/test_frontend_state_regressions.py
+++ b/tests/test_frontend_state_regressions.py
@@ -16,6 +16,7 @@ missing guard — so pinning the line is what actually stops the regression.
   which is undefined over plain HTTP — a silent no-op on a token shown once.
 """
 
+import re
 from pathlib import Path
 
 import pytest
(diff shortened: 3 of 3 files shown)


fabriziosalmi:
Fair, and applied — the brittleness was real, not stylistic. Added `_js_function_body`, which extracts a function by counting braces from its opening one rather than slicing to a hard-coded `"\n    }"`, and switched the line assertions to whitespace-tolerant regexes. Also applied it to the #491 guard in the same file, which had the same weakness in the form of a fixed 400-character lookbehind window.

Checked both directions rather than assuming:

- removing `if (!reissueEditingDomain) return;` still fails, with the intended message
- rewriting it as `if ( ! reissueEditingDomain ) {\n    return;\n}` keeps the suite green

So it still pins the behaviour and no longer pins the formatting. Worth noting these guards remain source-level by design — the file's own docstring explains why: the behaviours are only fully exercised by the Playwright suite, which needs a running container, and each defect was a missing line rather than a subtle interaction.

PR #496, opened 2026-08-02
fix: an in-flight issuance survives a page refresh (#399)

### LINKED ISSUE #399: when issuing a new cert. the new "issuing entry" that appears in the list dissapears if you refresh the page
**Describe the bug**
when issuing a new cert. the new "issuing entry" that appears in the cert list disappears if you refresh the page. after x amounts of minutes in the background the cert gets issued and the actual entrys gets added to the list

**To Reproduce**
start issuing a cert. 
refresh the page. 
the entry that for the new cert which was at the top of the list with the "issuing" status dissapears.

**Expected behavior**
issuing cert should still show on the list when page is refreshed or a new user session is opened on another browser

**Additional context**
it makes the user think the issue has stopped or failed since theres no longer an entry in the list showing the new cert thats being requested

fabriziosalmi:
Closes #399. Reported by @ITJamie.

## The bug

Start issuing a certificate, refresh the page, and the "Issuing" row at the top of the list disappears. The certificate does arrive minutes later — but in the meantime it reads as if the request had stopped or failed, which is exactly what the reporter said it looked like.

## Cause

The row was built entirely from `pendingJobs`, a plain object living in the page's memory. A reload lost it.

The server has tracked these jobs all along — `IssuanceExecutor` records `queued` / `running` / terminal — but exposed only `GET /api/certificates/jobs/<id>`. With no way to ask *what is in flight*, a freshly loaded page had nothing to rediscover, no matter how it was written.

## The fix

- **`IssuanceExecutor.list_active()`** returns queued and running jobs, as copies. Terminal jobs are excluded deliberately: a finished certificate is already in the list, and replaying it would show two rows for one domain.
- **`GET /api/certificates/jobs`** — operator-gated, like the single-job poll. A scoped API key gets its jobs **filtered** rather than a 403, so a restricted key still sees its own work and nobody else's.
- **The dashboard** calls it after the first render, re-attaches to whatever is in flight and resumes polling.

That last point also satisfies the half of the report that client-side persistence could never have covered:

> issuing cert should still show on the list when page is refreshed **or a new user session is opened on another browser**

Because the state is the server's, a second browser sees the same in-flight row.

## One thing the fix had to be careful about

A job adopted after a refresh has **no original request body** — this page never saw the form. `retryCreateJob` does `job.payload || {}`, so a Retry button on such a row would have POSTed an empty create. `failedRowHtml` now omits Retry when there is no payload; Dismiss is unaffected.

## Tests

- `IssuanceExecutor.list_active()`: lists in-flight only, excludes finished, and returns copies (mutating the result must not corrupt the registry).
- The endpoint: lists in-flight jobs, filters by domain scope for a scoped key, and 404s when async issuance is disabled.
- A source-level guard for the dashboard rehydration, in the established pattern of `test_frontend_state_regressions.py`, including the missing-payload Retry case.

Full suite: **2241 passed**, 6 skipped. flake8, bandit and `theme_codemod --check` clean.

### DIFF
--- a/modules/api/resources.py
+++ b/modules/api/resources.py
@@ -2477,6 +2477,33 @@ def create_api_resources(api, models, managers):
                 return scope_err
             return job, 200
 
+    class CertificateJobs(Resource):
+        @api.doc(security='Bearer')
+        @auth_manager.require_role('operator')
+        def get(self):
+            """List the issuance/renewal jobs still in flight.
+
+            The dashboard builds its "issuing" row from client-side state, so
+            a refresh made an in-flight issuance disappear from the list and
+            look like it had failed (issue #399). Listing the jobs the server
+            already tracks lets any session rediscover them — including one
+            opened in a different browser.
+
+            Only queued/running jobs are returned; a finished one is already
+            represented by the certificate itself.
+            """
+            if cert_executor is None:
+                return {'error': 'Async issuance is not enabled'}, 404
+            # Same boundary as polling a single job: a scoped key sees only
+            # jobs for domains it could have created itself. Filtered rather
+            # than refused, so a scoped key still gets its own in-flight work.
+            user = getattr(request, 'current_user', None) or {}
+            jobs = [
+                job for job in cert_executor.list_active()
+                if auth_manager.user_can_access_domain(user, job.get('domain'))
+            ]
+            return {'jobs': jobs, 'count': len(jobs)}, 200
+
     class CertificateAutoRenew(Resource):
         @api.doc(security='Bearer')
         @auth_manager.require_role('operator')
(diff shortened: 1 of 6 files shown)


fabriziosalmi:
You were right, and the comment was the actual defect: it asserted a guarantee that did not exist. `retryCreateJob` does `job.payload || {}` and the only protection was `failedRowHtml` hiding the button — precisely the mismatch that gets a guard deleted later by someone who read the comment and believed it.

Made the claim true rather than softening it. `retryCreateJob` is exposed on `window` and reachable without the button, so it now refuses a payload-less job with an explanation instead of resubmitting an empty create, and the row still omits Retry. Two independent protections, both pinned by the regression guard.

Issue #456, opened 2026-07-22
mcp/: seven Dependabot alerts in the MCP server's transitive npm tree


fabriziosalmi:
Seven Dependabot alerts opened today, all of them in `mcp/package-lock.json`: `fast-uri` (2 high), `hono` (3 moderate), `@hono/node-server` (moderate), `body-parser` (low).

## Scope, first, because the headline number is misleading

All seven are **transitive** dependencies of `@modelcontextprotocol/sdk`. `mcp/package.json` declares two direct dependencies:

```json
"dependencies": { "@modelcontextprotocol/sdk": "^1.0.1", "node-fetch": "^2.7.0" }
```

They do **not** affect the published Docker image at runtime. The image ships the `mcp/` sources (`index.js`, `package.json`, `package-lock.json`) but no `node_modules` and no Node runtime, so the vulnerable code is not present in it — only the lockfile that names it. Anyone who runs the MCP server for real does `npm install` outside the image, and they are the ones exposed.

## What to do

- Bump `@modelcontextprotocol/sdk` to a version whose tree resolves past all four packages, refresh the lockfile, and check the server still starts and answers a tool call.
- If the SDK has not moved, `overrides` in `mcp/package.json` is the honest interim, with a comment naming the advisories it pins past.
- Separately worth deciding: `mcp/` is copied into the production image at all. #426 removed `tests/`, `docs/` and `scratch/` from it for the same reason — the image should carry what it runs. That is a one-line `.dockerignore` change, but it is a behaviour change for anyone who copies the server out of the image, so it wants its own decision rather than a drive-by.

Not blocking v2.23.0: the release artifact does not execute any of this.

fabriziosalmi:
Update after the 2026-07-28 draconian sweep. The alert count went from 7 to 1.

Landed (all mcp/CI hygiene, none of it in the runtime image):
- #463 fast-uri 3.1.2→3.1.4 — closes the 2 **high** fast-uri alerts
- #464 hono 4.12.25→4.12.32 — closes the 3 **moderate** hono alerts
- #453 body-parser 2.2.2→2.3.0 — closes the **low** body-parser alert
- #462 postcss 8.5.10→8.5.23 — closes 2 **high** postcss alerts that surfaced in the root dev toolchain
- #447 base-image digest bump (helps #403) and #461 the actions group, for good measure

[...]

PR #499, opened 2026-08-02
fix(mcp): resolve @hono/node-server to 2.0.12 (GHSA-frvp-7c67-39w9)

fabriziosalmi:
The last of the seven Dependabot alerts tracked in #456. With this, that issue can close.

## Why it was stuck

`@hono/node-server` is **transitive** — nothing in `mcp/package.json` asks for it; it arrives via `@modelcontextprotocol/sdk`. That usually means waiting for the upstream package to move.

Not here. The SDK already declares:

```json
"@hono/node-server": "^1.19.9 || ^2.0.5"
```

The patched line was permitted all along; the lockfile had simply resolved to 1.19.14 and never moved. So `npm update @hono/node-server --package-lock-only` lands on 2.0.12 **inside the range the SDK itself asks for** — no `overrides` block, no constraint bent, nothing forced on a dependency we do not own.

## The diff

Lockfile only: four lines, one package, nothing added or removed.

```
- "version": "1.19.14"
+ "version": "2.0.12"
```

Run with `node:20` in Docker to match CI rather than whatever npm happens to be on the machine.

## On actual exposure

Worth stating plainly rather than implying urgency: the advisory is a **path traversal in `serve-static` on Windows**, and this MCP server never calls `serveStatic` — verified by grep across its source. So this is hygiene, not an incident.

It is still worth doing. An alert that nobody can close is exactly how a real one gets skimmed past later, which is the same argument the release notes for #488 made about noisy version drift.

## After merge

#456 (`mcp/: seven Dependabot alerts in the MCP server's transitive npm tree`) should go to zero open alerts and can be closed.


fabriziosalmi:
Verified and correct — thank you, this would have shipped a real inconsistency.

```
@hono/node-server 2.0.12  engines: node >=20
@hono/node-server 1.19.14 engines: node >=18.14.1
mcp/package.json          engines: node >=18.0.0
```

Reverting was not an option: the advisory covers everything below 2.0.5, so there is no fixed 1.x line to fall back to. The choice was between keeping an inaccurate support claim and correcting it.

Corrected it. `mcp/package.json` now declares `>=20.0.0`, and the lockfile root was refreshed to match.

That is honest rather than a concession: **Node 18 has been end-of-life since April 2025**, CI already runs Node 20 (`.github/workflows/ci.yml`), and nothing else in the repository claims 18. This one field was the last thing saying otherwise — it was already wrong before the bump made it visible.

Issue #233, opened 2026-05-20
[Feature] RSA Key types.


tuxpowered:
I noticed that even when selecting RSA key while creating a cert, the format is the newer format. This is fine, and works 99% of the time.

However there are instances where systems do not yet accept the newer RSA key and what the older style keys.

PKCS#8 -> PKCS#1 

Would it be possible to have the key saved in different formats? This sort of overlaps with the concept of #230

PR #500, opened 2026-08-02
feat(api): allow key_format with format=json on certificate download (#398)

### LINKED ISSUE #398: [Feature] Enhance API for RSA Key
The ability to `download` the private key in RSA format was resolved in #237. However as the parameters onlly allow that to be downloaded. It is not available on the API endpoing, Meaning the file needs to be saved and then read into the workflow again.

The download is fine, but can we enhance the `download` api? 

Currently `key_format=pkcs1` is locked to requiring `file=privkey.pem` It would be usefull to allow the key_format to be included with the `format=json`

```bash
 curl -H "Authorization: Bearer  shhhh"  "http://10.0.0.3:8000/api/certificates/dfatech.ca/download?format=json&key_format=pkcs1`"
```
I would propose that if `key_format=pkcs1` is passed in because it currently is converted from the private_key_pem, that we continue to include the original key in the json output but add `private_key_rsa_pem` to the json output.

**Additional context**
adding the URL param to key_format=pkcs1 allows pulling everything in 1 API call to facilitate api automations.

fabriziosalmi:
Closes #398. Requested by @tuxpowered.

## The gap

`key_format=pkcs1` shipped in #233 to convert certbot's PKCS#8 key into the legacy traditional form, but the endpoint rejected it unless paired with `?file=privkey.pem`:

```python
if key_format and requested_file != 'privkey.pem':
    return {'error': 'key_format only applies to ?file=privkey.pem'}, 400
```

So an automation that wanted the whole bundle *and* the legacy key had to make a second call and stage the key through a file on disk — exactly the workflow described in the request.

## The change

`?format=json&key_format=pkcs1` now returns the converted key inline.

Crucially it is **added, not substituted**: `private_key_pem` stays byte-for-byte what it was, so nothing that already consumes `format=json` changes. That was the reporter's own proposal and it is the right call. `key_format=pkcs8` on the JSON form is a no-op, since PKCS#8 is what is on disk.

The conversion runs on the bytes already read for the payload — no second file read, and no second path to validate.

## One deviation from the request, and why

The request proposed the field be called `private_key_rsa_pem`. It is named **`private_key_pkcs1_pem`** instead.

The traditional/PKCS#1 encoding only produces `BEGIN RSA PRIVATE KEY` for RSA keys. For ECDSA it produces SEC1, `BEGIN EC PRIVATE KEY` — which this repository already pins in `tests/test_privkey_pkcs1_export.py::test_ecdsa_pkcs8_converts_to_sec1` — and **CertMate issues ECDSA by default**. A field with `rsa` in its name would therefore contain an EC key for most users, and a public API field name is hard to walk back later.

Naming it for the encoding matches the `key_format=pkcs1` parameter that already exists. There is a test asserting exactly this, so the reasoning is pinned rather than left in a commit message.

## Tests

`tests/test_issue398_json_key_format.py`, in-process, no Docker:

- pkcs1 is added and `private_key_pem` is untouched, rest of the bundle unaffected
- absent by default — the previous response shape is unchanged
- `pkcs8` is a no-op
- an invalid `key_format` is still a 400
- ECDSA yields SEC1 (the test that documents the naming decision)
- Ed25519 returns **422**, not a 500
- `key_format` with a *public* file is still rejected — relaxing the guard for JSON must not relax it for files

Full suite: **2249 passed**, 6 skipped. flake8 and bandit clean. `docs/api.md` updated.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -509,13 +509,17 @@ Download certificate files for a specific domain. By default, this endpoint retu
 - `file` (Query, Optional) - Specify a single file to download. 
   - Supported values: `fullchain.pem`, `privkey.pem`, `combined.pem`
 - `format` (Query, Optional) - Set to `json` to return all certificate files in a JSON object.
+- `key_format` (Query, Optional) - `pkcs1` or `pkcs8`. Certbot writes PKCS#8 (`BEGIN PRIVATE KEY`); some older stacks require the legacy traditional form. Valid with `file=privkey.pem` (serves the converted key) or with `format=json` (adds a converted copy to the response).
 
 **Response** (200 OK):
 - **Default**: `application/zip` (A ZIP file containing all PEM files)
 - **With `file` param**: `application/x-pem-file` (The raw content of the requested file)
 - **With `format=json`**: `application/json` with `domain`, `cert_pem`, `chain_pem`, `fullchain_pem`, and `private_key_pem`
+- **With `format=json&key_format=pkcs1`**: the above plus `private_key_pkcs1_pem`
 
 The JSON form is the preferred automation shape for Ansible, Salt, or any other client that wants to write PEM files directly.
+
+`key_format=pkcs1` on the JSON form **adds** `private_key_pkcs1_pem` and leaves `private_key_pem` untouched, so an existing consumer is unaffected and a client needing the legacy key no longer has to make a second call and stage it through a file. The field is named for the encoding rather than for RSA: the traditional form of an ECDSA key is SEC1 (`BEGIN EC PRIVATE KEY`), and CertMate issues ECDSA by default. Key types with no traditional encoding (Ed25519) return **422**.
 
 **Examples**:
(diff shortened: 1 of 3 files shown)


fabriziosalmi:
Applied, and it was worth doing — the three failures were indistinguishable in the logs while needing completely different responses from an operator.

I did check before widening it, since this is an error path on a private key. `cryptography` emits static descriptive strings with no key material:

```
ed25519      -> ValueError | format is invalid with this key
encrypted    -> TypeError  | Password was not given but private key is encrypted
malformed    -> ValueError | Could not deserialize key data. The data may be in ...
```

So the message is safe to log and genuinely diagnostic. CR/LF scrubbed like every other interpolated value in this module.

Applied to **both** conversion sites — the new `format=json` one and the pre-existing `?file=privkey.pem` one, which had the same limitation. Fixing only the new one would have left them to drift.

PR #501, opened 2026-08-02
feat(cli): certmate cert download, so a host can pull what it deploys (#490)

### LINKED ISSUE #490: [Feature] Support downloading certificates via certmate-cli
When deploying certificates to remote servers it is often easier for the remote server to pull the certificates it needs rather than have them pushed from the certificate manager for two reasons: firstly, certificate deployment may require steps that cannot easily be performed remotely (e.g. restarting services or running application specific commands on the remote server); secondly, pushing certificates to multiple remote servers requires that the central server have credentials to access all of the remote servers that must be deployed to, whereas pulling certificates only requires that the remote servers have access to pull the certificates they need.

Add the ability to download certificates via a command such as `certmate cert download domain`, potentially with arguments to specify which file(s)

Using API directly via `wget` or `curl`

**Additional context**
I expected that the certmate-cli tool would be able to fetch certificates managed by CertMate out of the box, as the ability to request a certificate via the CLI is not very useful if the certificate itself is not available

fabriziosalmi:
Closes #490. Requested by @exterrestris.

## The argument for it

Straight from the request, and it is the right one: pushing means the certificate manager holds credentials for **every** target host; pulling means each host holds **one** credential, for itself.

## The gap was narrower than it looked

Every layer had this except the one you would actually use:

| layer | before |
|---|---|
| server API | `?file=`, `?format=json`, path-style `/download/<name>` — all present |
| `certmate-sdk` | `download_certificate()` for json/pem/zip/pfx |
| `certmate-cli` | **no download command at all** |

## Per-file first

```bash
certmate cert download example.com --file fullchain -o /etc/ssl/certs/x.pem
certmate cert download example.com --file privkey   -o /etc/ssl/private/x.key
```

with `--bundle zip|json` for whole-certificate shapes and `-o -` for stdout.

Combined with scoped API keys, each host gets a key limited to its own domain and pulls on a timer: no inbound access to the host, no credentials for it on the manager. The role split already on the endpoint fits — `cert`/`chain`/`fullchain` are viewer-readable, `privkey`/`combined`/`pfx` need operator.

## Details that matter more than the command

- **Files are created 0600 via `os.open` with the mode**, not `chmod`-ed after writing. Between an `open()` and a `chmod()` a private key is readable by anyone on the box — a small window, but this is the one file where it matters.
- **The success line reports the mode actually on disk.** `O_CREAT` leaves an existing file's mode alone, so printing `0600` unconditionally would sometimes be false.
- **The SDK sends `?file=<name>`**, not the path-style `/download/<name>`. Both exist server-side, but only the query-string form accepts `key_format` — so `--file privkey --key-format pkcs1` also serves the legacy key (#398, #233) through the same call.
- **An unknown `--file` fails before any request is sent.**

## Versioning

Clients bumped to **0.1.3**, and the CLI's floor on `certmate-sdk` raised to `>=0.1.3`. Left at `>=0.1.2`, installing the new CLI could resolve SDK 0.1.2 and fail at runtime on the missing method — the sort of thing that only shows up for the user, never in CI.

Publishing is a separate cycle (`clients-v*` tag), so this PR only prepares it.

## Tests

- **SDK**: short name → `?file=` mapping for all six files, `key_format` forwarding, and an unknown name raising before any request.
- **CLI**: writes the requested file, asserts the resulting mode is **0600**, defaults to `fullchain`, forwards `--key-format`, rejects an unknown `--file` without calling the server, rejects `--key-format` on a bundle, and writes a JSON bundle as JSON.

Full suite: **2263 passed**, 6 skipped. CLI README documents the pull-based pattern.

### DIFF
--- a/clients/certmate-cli/README.md
+++ b/clients/certmate-cli/README.md
@@ -16,6 +16,30 @@ certmate cert create app.example.com --dns cloudflare --dry-run
 certmate audit verify
 ```
 
+## Pulling certificates onto a host
+
+`cert download` fetches one file at a time, so a target server can pull
+exactly what it deploys instead of the certificate manager pushing to it:
+
+```bash
+certmate cert download app.example.com --file fullchain -o /etc/ssl/certs/app.pem
+certmate cert download app.example.com --file privkey   -o /etc/ssl/private/app.key
+```
+
+Files are created **0600**, with the mode set at creation rather than after
+the write, so the key is never briefly world-readable.
+
+This is worth preferring over pushing when the manager would otherwise need
+credentials on every target host. Give each host an API key scoped to its own
+domain and run the pull on a timer: the host needs no inbound access, and the
+manager holds no credentials for it. `cert`, `chain` and `fullchain` are
+readable by a viewer-role key; `privkey`, `combined` and `pfx` need operator.
+
+`--file privkey --key-format pkcs1` serves the legacy
+`BEGIN RSA PRIVATE KEY` form for stacks that reject certbot's PKCS#8.
+`--bundle zip` or `--bundle json` fetch the whole certificate instead, and
+`-o -` writes to stdout.
+
 Connection comes from `--url`/`--token` or `CERTMATE_URL`/`CERTMATE_TOKEN`.
 Prefer the `CERTMATE_TOKEN` environment variable over `--token`: command-line
 arguments are visible to other local processes (`ps`) and shell history.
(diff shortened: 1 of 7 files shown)


fabriziosalmi:
Both applied — and it was an inconsistency in my own code rather than a difference of opinion: the file **name** was already validated before any request went out, while `key_format` was documented and then forwarded unchecked. Same function, two standards.

Both layers now enforce what they document:

- SDK raises `ValueError` for a `key_format` that is not `pkcs1`/`pkcs8`, and for one passed with anything other than `privkey`
- CLI fails with the same two messages before constructing a request

A round trip to be told "invalid key_format" is a worse error than one raised locally, and it costs an authenticated call to learn it. Tests added on both sides asserting no request is sent.

Issue #475, opened 2026-07-28
Typed deploy target: write to a Kubernetes Secret


fabriziosalmi:
## Problem
Deploy hooks today are raw shell commands only. Writing a renewed certificate into a Kubernetes TLS Secret — one of the most common targets — means hand-rolling a shell hook with `kubectl` and credentials.

## Proposal
- Introduce a small **typed deploy-target** abstraction alongside the existing shell hooks.
- First typed target: **Kubernetes Secret** — write `tls.crt` / `tls.key` to a named Secret via kubeconfig or in-cluster credentials, on issuance and every renewal.
- Keep credentials in the existing storage-backend mechanism.

Scope is limited to typed deploy targets; no agent/fleet delegation.

## Acceptance
- A managed certificate can be configured to publish to a Kubernetes Secret and does so automatically on renewal (verified against a mocked API).

fabriziosalmi:
Done — merged in #480 (M5).

PR #504, opened 2026-08-02
feat(k8s): Helm chart for running CertMate in a cluster

fabriziosalmi:
Closes the gap identified in the distribution review: CertMate ships a **Kubernetes Secret deploy target** (#475) and a guide for getting certificates *into* a cluster, but no way to install CertMate *itself* there short of hand-translating `docker-compose.yml`. Kubernetes users install with `helm`. Without a chart, the project is not evaluated in the environment where its closest comparison lives.

```bash
helm install certmate ./charts/certmate --namespace certmate --create-namespace
```

## The chart encodes what CertMate is, rather than emitting boilerplate

**One replica, and it refuses to render more.** APScheduler runs inside the web process and gunicorn runs a single worker on purpose. `--set replicaCount=2` is not a scaled CertMate — it is two schedulers issuing and renewing against the same certificate store, and two writers on a ReadWriteOnce volume. That fails at template time with the reason:

```
Error: certmate: replicaCount must be 1. CertMate is single-instance by design —
its scheduler runs in-process, so a second replica duplicates every renewal job
against the same certificate store. Scale up the resources, not the replicas.
```

A duplicate-issuance deployment looks perfectly healthy in `kubectl`. Template time is the only place to catch it before it is someone's incident.

**`Recreate`, not `RollingUpdate`.** The new pod cannot mount a ReadWriteOnce volume while the old pod holds it, and the old pod is not terminated until the new one is Ready — a rolling update deadlocks. A few seconds of downtime is the right trade here.

**It will not render without storage.** Certificates, the settings store, the inventory and the tamper-evident audit chain are not pod-local state. `persistence.enabled=false` fails unless `persistence.existingClaim` is supplied — the escape hatch for operators who manage volumes themselves.

**The volume outlives the release.** `helm.sh/resource-policy: keep` on the PVC, so `helm uninstall` does not take the certificates with it.

**One claim, four subPaths.** Four separate RWO claims would buy nothing on a single-instance workload.

**`existingSecret` support**, because `values.yaml` is not a secret store, and a checksum annotation so a rotated token actually restarts the process instead of sitting unread in the API server.

## The version string is wired in from day one

`appVersion` is another copy of the release number. Left alone it becomes the next stale one — that lesson has already cost two hand-fixes for the docs landing pages (#483, #487) and one for `package-lock.json` (#498). So it is bumped by `scripts/release.sh` and pinned by `test_version_consistency` in the same commit that introduces it.

## Tests

`tests/test_helm_chart.py`:

- **Static, always run**: replicas are hardcoded and not templated, the render-time guard exists, the strategy is `Recreate`, the PVC keeps its data, `appVersion` matches `modules.__version__`.
- **Render, when `helm` is present**: defaults render, both guards fire with their intended messages, and `existingClaim` is accepted while suppressing the PVC.

The render tests **skip where helm is not installed**, which includes CI today. Stated rather than papered over: the invariants that matter most are covered by the static half, which runs everywhere. Adding `azure/setup-helm` to the CI workflow would make the render half real too — happy to do that as a follow-up if you want it.

Full suite: **2278 passed**, 6 skipped. `helm lint` clean.

### DIFF
--- a/charts/certmate/Chart.yaml
+++ b/charts/certmate/Chart.yaml
@@ -1,0 +1,25 @@
+apiVersion: v2
+name: certmate
+description: Self-hosted SSL/TLS certificate lifecycle management — issuance, renewal, discovery, inventory and deployment
+type: application
+# Chart version: bumped when the CHART changes.
+version: 0.1.0
+# Application version: tracks the CertMate release this chart defaults to.
+# scripts/release.sh bumps this, and tests/test_version_consistency.py fails
+# the build if it drifts from modules.__version__ — the same treatment the
+# docs landing pages and package-lock.json needed.
+appVersion: "2.25.0"
+home: https://www.certmate.org
+sources:
+  - https://github.com/fabriziosalmi/certmate
+keywords:
+  - certificates
+  - tls
+  - ssl
+  - acme
+  - lets-encrypt
+  - certificate-lifecycle-management
+  - private-ca
+maintainers:
+  - name: fabriziosalmi
+    url: https://github.com/fabriziosalmi
--- a/charts/certmate/README.md
+++ b/charts/certmate/README.md
@@ -1,0 +1,82 @@
+# CertMate Helm chart
+
+Run [CertMate](https://github.com/fabriziosalmi/certmate) in Kubernetes.
+
+```bash
+helm install certmate ./charts/certmate --namespace certmate --create-namespace
+kubectl -n certmate port-forward svc/certmate 8000:8000
+```
+
+## What this chart deliberately will not do
+
+**It renders exactly one replica, and refuses anything else.** CertMate's
+scheduler (APScheduler) runs inside the web process and gunicorn runs a single
+worker on purpose. A second replica is a second scheduler issuing and renewing
+against the same certificate store, and a second writer on a ReadWriteOnce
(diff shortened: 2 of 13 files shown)


fabriziosalmi:
Both applied, and both landed on the path the chart itself recommends — which makes them worth more than they look.

**`optional: true` on the secretRef.** Removed. A Secret that is *named* must exist; a typo in `existingSecret` now stops the pod with a clear reason instead of starting it without credentials.

That was only safe once the reason it was there was fixed: with no secret values the chart still rendered a Secret with an empty `stringData:` and pointed `envFrom` at it. It now generates a Secret **only when there is something to put in it**, and omits `envFrom` entirely otherwise — CertMate generates its own token on first boot, which is the documented trial path. Three cases, each rendering the honest thing, each pinned by a test.

**The checksum not covering `existingSecret`.** Correct, and I could not fix it without making the chart worse. Reading a Secret the chart does not own requires `lookup`, which returns nothing during `helm template` and makes rendering depend on the live cluster — that breaks dry runs and GitOps diffs, which is a bigger cost than the gap.

So it is documented rather than papered over: the README now says plainly that rotating an external Secret needs `kubectl rollout restart` or a controller like [Reloader](https://github.com/stakater/Reloader), and the template carries the same reasoning in a comment so nobody "fixes" it with `lookup` later.

PR #505, opened 2026-08-02
docs: describe CertMate as what it became, not what it started as

fabriziosalmi:
The repository description and README opening still described v1: *"an SSL certificate management system … multi-DNS provider support, Docker containerization, and a comprehensive REST API"*.

That is an accurate description of an ACME issuer with a web UI, which places CertMate in a comparison with `acme.sh` — a comparison it cannot win on features it stopped competing on two milestones ago.

## What it actually does now

Since v2.24.0, on top of issuance and renewal:

- **discovers** certificates it did not issue, by probing hosts and by watching CT logs for shadow issuance
- keeps a **fingerprint-keyed inventory** of what is actually served across the estate — where, by whom, expiring when, using which cryptography
- reports **cryptographic readiness** across everything it knows about
- **deploys** renewals to typed targets, not just shell hooks
- runs a **private CA** with CRLs and mTLS, behind a tamper-evident audit trail with SIEM export

That is certificate lifecycle management. Different search term, different evaluation, different reader.

## Why it matters more than it looks

The description is what GitHub search ranks, what Google shows, and what every shared link preview renders. It is the single highest-leverage sentence in the repository, and it had not been touched in two milestones of work.

This PR updates the README title and opening paragraph. The GitHub repository description and topics are set outside the repo and are being updated alongside it.

No functional change; full suite green (2282 passed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 <img src="certmate_logo.png" alt="CertMate Logo" width="180">
 
-**CertMate** is a self-hosted certificate lifecycle management platform: it issues and renews TLS certificates, **discovers the ones you did not issue**, keeps a single inventory of what exists across your estate — what is served where, who issued it, when it expires, which cryptography it uses — and deploys renewed certificates to where they are needed. Multi-DNS provider support, a private CA, a tamper-evident audit trail, and a REST API for all of it.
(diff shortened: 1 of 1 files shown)


fabriziosalmi:
Fair, and the rewrite is better than the fix: 'supports / runs / keeps / exposes' states what the software *does*, where a trailing noun list only named parts.

Took the provider count from `DNSManager.SUPPORTED_PROVIDERS` — **29** — rather than a rounded "30+" that would drift the moment one is added or removed. Same reasoning as everything else in this repository that quotes a number.

PR #506, opened 2026-08-02
feat(ci): publish the Helm chart to GHCR and announce releases

fabriziosalmi:
Two distribution gaps from the review, both automated rather than left to memory.

## 1. The chart is actually published

`helm install ./charts/certmate` only works from a checkout, which is half a chart. On every published release it now goes to GHCR as an OCI artifact:

```bash
helm install certmate oci://ghcr.io/fabriziosalmi/charts/certmate --version 2.25.0
```

**GHCR rather than Docker Hub**, for a concrete reason: the container image is at `docker.io/fabriziosalmi/certmate`, and Docker Hub has no nested namespaces on that account. Pushing the chart there would put a chart and an image in **one OCI repository fighting over the same tags** — `2.25.0` would mean two different artifacts. GHCR supports nested paths, so the chart lands under `charts/` and cannot collide. `GITHUB_TOKEN` also covers it, so this adds no secret.

No `helm repo add`, no `gh-pages` branch, no `index.yaml` to keep in sync — Helm 3.8+ pulls `oci://` directly.

### Gated like the client publish

An OCI push is effectively write-once per tag, so the workflow **refuses to publish a chart whose version disagrees with the release tag**, mirroring `publish-clients.yml`'s check against the pyproject versions.

It also re-runs the two render guards in CI:

```
helm template --set replicaCount=2         -> must FAIL
helm template --set persistence.enabled=false -> must FAIL
```

That is deliberate duplication. The pytest versions skip wherever `helm` is absent, and CI is exactly where it is absent — a guard that only runs on my laptop is not a guard.

## 2. The chart versions with the application

`version` and `appVersion` are both the CertMate version, bumped together by `release.sh`, pinned by tests.

A separate chart version is a second number to bump, and **every independent copy of a version in this repository has drifted at least once**: the docs landing pages twice (#483, #487), `package-lock.json` two releases behind (#498). `version` also *has* to move for a publish to be a new artifact rather than a rejected duplicate — a chart whose contents changed under an unchanged version is how a stale chart gets served forever.

A chart-only fix rides the next release. That is the cost, and it is smaller than the drift.

## 3. Releases are announced

**1,240 stars, 7 watchers.** Shipping a release notifies almost nobody — four versions went out in two days and the people whose bugs were fixed only found out because they were told individually.

A Discussion in Announcements is the one surface people can subscribe to, and it shows on the repository front page. Now posted automatically on release, carrying the notes, the `docker pull` and `helm upgrade` commands, and a link back.

Two details: it reads the release **through the API rather than the event payload**, so a manual re-run announces exactly what an automatic one would; and it **skips when a discussion with that title already exists**, because a duplicate announcement is worse than none.

## Verification

`helm lint` clean, `helm package` produces `certmate-2.25.0.tgz`, both workflow files parse, full suite **2283 passed**.

The publish workflow can be exercised without shipping anything via `workflow_dispatch` with `dry_run: true`, which packages and runs every gate but does not log in or push.

### DIFF
--- a/.github/workflows/announce-release.yml
+++ b/.github/workflows/announce-release.yml
@@ -37,19 +37,34 @@ jobs:
 
           # Read the release rather than trusting the event payload, so a
           # manual re-run announces the same thing an automatic one would.
-          name="$(gh release view "$TAG" --repo "$REPO" --json name -q .name)"
-          body="$(gh release view "$TAG" --repo "$REPO" --json body -q .body)"
+          name="$(gh release view "$TAG" --repo "$REPO" --json name -q '.name // ""')"
+          body="$(gh release view "$TAG" --repo "$REPO" --json body -q '.body // ""')"
           url="$(gh release view "$TAG" --repo "$REPO" --json url -q .url)"
           version="${TAG#v}"
 
+          # A release is not required to have a name, and an empty title would
+          # create an untitled discussion AND break the de-duplication below.
+          # The tag is the one thing always present, so the title is keyed on
+          # it: either it already appears in the name, or it is prefixed.
+          if [ -z "$name" ]; then
+            title="$TAG"
+          else
+            case "$name" in
+              *"$TAG"*) title="$name" ;;
+              *) title="$TAG - $name" ;;
+            esac
+          fi
+
           # Skip if this release was already announced. Re-running the workflow
-          # should be safe: a duplicate announcement is worse than none.
+          # should be safe: a duplicate announcement is worse than none. Keyed
+          # on the tag rather than the full title, which is stable even if the
+          # release is later renamed.
           existing="$(gh api graphql -f query='
             query($q: String!) {
               search(query: $q, type: DISCUSSION, first: 5) {
                 nodes { ... on Discussion { title url } }
               }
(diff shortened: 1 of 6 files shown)


fabriziosalmi:
All four applied. The first was the worst of them, and it was mine.

**1. The contradictory comment.** `release.sh` still said the chart *version* is deliberately not bumped, directly above the code that now bumps it. A comment that contradicts its own code is worse than no comment — it is what the next person believes. Rewritten to state the actual policy.

**2. Releases with no name.** Correct, and it broke two things at once: an empty title, and de-duplication that searched on that title. The title is now keyed on the **tag**:

```
name='v2.25.0 (Pulling certificates)' -> title=v2.25.0 (Pulling certificates)
name='Pulling certificates'           -> title=v2.25.0 - Pulling certificates
name=''                               -> title=v2.25.0
```

and the duplicate check searches the tag, which is stable even if the release is renamed later.

**3 & 4. Asymmetric quote stripping.** A valid `version: "2.25.0"` would have failed the equality gate it exists to enforce, and printed quoted into a copy-pasteable helm command. Both now strip. Verified by quoting the field and reading it back:

```
$ helm show chart charts/certmate | awk '/^version:/{gsub(/"/,"",$2); print $2}'
2.25.0
```

Suite still green at 2283.

PR #507, opened 2026-08-02
chore(ci): announcements are a manual step, not an automatic one

fabriziosalmi:
Follow-up to #506, on the maintainer's reservation about automatic announcements — which is well placed.

## What changes

`announce-release.yml` no longer triggers on `release: published`. It is `workflow_dispatch` only:

```bash
gh workflow run announce-release.yml -f tag=v2.25.0
```

## Why this one is manual while the chart publish is not

Both were added in the same PR and it is worth being explicit about why they now differ.

**Publishing an artifact is mechanical.** The chart either matches the release tag or the gate refuses to push. Nothing subjective happens, and a wrong push is fixable by pushing a corrected version.

**An announcement is the maintainer's voice on a public forum, and once posted it has been read.** Firing it automatically means a wrong release note, a mis-tagged release, or a release cut and then withdrawn all become communication before anyone has looked at them. A Discussion cannot be un-sent the way a tag can be moved.

The failure modes are not symmetric, so the automation should not be either.

## What is kept

Everything that made it worth writing: the tag-keyed title that survives an unnamed release, the duplicate check so a re-run is safe, and the composed body with the `docker pull` and `helm upgrade` commands and a link back to the notes.

It stops being an automation and becomes a tool — which still solves the original problem (7 watchers, releases nobody hears about), just with a human deciding when.

### DIFF
--- a/.github/workflows/announce-release.yml
+++ b/.github/workflows/announce-release.yml
@@ -46,11 +46,21 @@ jobs:
         run: |
           set -euo pipefail
 
+          # TAG is a workflow_dispatch input. Validate it before it reaches a
+          # command line or a query string: a value starting with '-' would be
+          # read as a flag by gh, and quotes or a newline would break out of
+          # the GraphQL search string below. `case` compares the whole value —
+          # grep would match line by line and let a newline through.
+          case "$TAG" in
+            "" | -*)               echo "::error::tag must be non-empty and must not start with '-': $TAG" >&2; exit 1 ;;
+            *[!A-Za-z0-9._-]*)     echo "::error::tag may only contain letters, digits, dot, underscore and hyphen: $TAG" >&2; exit 1 ;;
+          esac
+
           # Read the release rather than trusting the event payload, so a
-          # manual re-run announces the same thing an automatic one would.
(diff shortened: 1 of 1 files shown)


fabriziosalmi:
Applied — and it caught a hole in my first attempt at the fix, which is worth recording.

The tag is validated before it reaches any command line or query string, and passed after `--` so it is unambiguously positional.

**I first wrote the check with `grep -Eq '^[A-Za-z0-9][A-Za-z0-9._-]*$'`, and it was wrong.** grep matches **line by line**, so a tag of `v1\nx` passed: the first line matched the pattern and grep returned 0. Exactly the injection vector the review was pointing at, waved through by the guard meant to stop it.

`case` compares the whole value, so it has no such gap. Verified against every case:

```
v2.25.0                -> OK
clients-v0.1.3         -> OK
--help                 -> REJECTED
-v1                    -> REJECTED
v1.0.0; rm -rf /       -> REJECTED
v1"x                   -> REJECTED
v1\nx                  -> REJECTED   (passed the grep version)
""                     -> REJECTED
v1 x                   -> REJECTED
```

A guard is only worth having once you have watched it fail on the thing it exists to stop.

PR #581, opened 2026-08-21
fix: the four defects that survived the full-field audit

fabriziosalmi:
A 146-agent audit ran over twelve dimensions of this codebase on 2026-08-18, with an adversarial layer whose job was to demolish each finding rather than confirm it. 66 findings were proposed, 23 were killed by the refuters, 43 survived — and every one of the survivors came back with a correction attached, most of them deflating the severity the finder had claimed.

I then re-read the worst ones in the code myself, because two agents agreeing is not evidence. Four held up. This PR closes those four and nothing else; the remaining 39 are smaller and follow separately, grouped by theme.

Each fix has a test that fails on the code it describes. That is the acceptance criterion for all four — no exceptions, because "it looks right now" is how three of these got in.

## 1. The settings lock protected a snapshot, not the file

`update()` and `atomic_update()` hold `self._lock` for the whole read-modify-write, so two writes cannot interleave. But the "read" was `load_settings()`, which inside a request returns the dict cached on `flask.g` when the request *started*.

The window is not milliseconds. On the synchronous issuance path — the default, since async is opt-in — `cert_service.py` loads settings at :183, runs certbot, and writes at :242, all in the same request thread. For the whole duration of an issuance every other write to `settings.json` was silently reverted: a user created, a domain registered, a DNS credential saved. **A domain rolled out of `settings['domains']` is never visited by `check_renewals` again and its certificate expires in silence.**

The same snapshot defeated the protection meant to prevent exactly this: `atomic_update`'s `protected_keys` restores `users` and `api_keys` from `existing`, so a cached `existing` restored a deleted account — password hash included — after `delete_user` had reported success.

`load_settings` grows `use_cache`; the two write methods pass `False`. Nothing outside `settings.py` calls `save_settings`, so those two methods are every write path in the product. The read path is untouched and all six contracts in `test_settings_request_scoped_cache.py` still hold.

## 2. A backup that cannot restore is not a backup

Every automatic backup is masked — `save_settings` calls `create_unified_backup(settings, reason)` and `include_secrets` defaults to `False`. That default is right: a leaked backup must not also be a credential dump. What was wrong is that every restore path treated those archives as authoritative. The manifest has recorded `secrets_masked` since unified backups existed; nothing read it.

So when `settings.json` could not be parsed, `load_settings` installed the newest masked archive as live settings: every password hash, bearer token, API key hash, OIDC client secret and DNS credential became the literal sentinel. No login, no token, no SSO, every renewal failing at the provider — and `setup_completed` still `True`, so the wizard did not reopen. The log said "Settings restored successfully from backup".

Four changes, one defect:

- the restore skips archives that cannot restore and picks the newest one that can, saying in the log what it found and how to make a usable backup;
- when nothing usable exists it **refuses instead of recreating the file**. `safe_file_read` returns the default for a JSON typo, a `PermissionError` and an empty read alike, and `settings.json` is written atomically, so the app never produces this state itself. Overwriting destroys the operator's only copy of a file a text editor could have repaired. `SettingsUnreadableError` deliberately escapes the catch-all below it, because returning defaults in-memory leaves `setup_completed` False — an instance that cannot read its credentials must not come up world-open;
- the pre-restore rollback archive is written with `include_secrets=True`. It exists solely to undo the restore that follows it; masked, it was a safety net that could not catch anything;
- **a login no longer writes a backup.** Recording `last_login` wrote a full ZIP of settings + certificates and then pruned to `MAX_BACKUPS_PER_TYPE`, so login traffic alone evicted every restore point an operator wanted. A password-hash upgrade still writes one.

`test_disaster_recovery_round_trip.py` is the test nobody had written: configure an instance, back it up, destroy it, restore it, log in. There were tests for creating backups, listing them, masking them, and the restore endpoint's authorization and zip-bomb limits — none that asked whether a backup can bring the instance back.

## 3. Setup mode must be a door that only closes

`_authenticate_request` returns `{'username': 'setup_user', 'role': 'admin'}` for a caller with no credential at all — enough to read `/api/settings`, mint API keys, create admins and download private keys.

v2.21.4 added the OIDC branch to `is_setup_mode()` to *close* a real hole: an SSO-only box has no bearer token and never gets `local_auth_enabled` set, so it sat world-open forever. That fix was right. Its side effect was that closure now depends on a checkbox — and unchecking "Enable OIDC/SSO" put the instance back into anonymous-admin, with no guard on the route and no warning beside the toggle.

`is_setup_mode()` is now expressed once as `setup_mode_for(settings)`, and `would_open_setup_mode(candidate)` asks the same question about a change not yet written. One definition, two callers: a guard that re-implements the predicate it defends drifts away from it. It returns `False` when the instance is already in setup mode, so a fresh install stays configurable.

On the OIDC route it fires only for the deliberate turn-off: a payload clearing `issuer_url` while `enabled` stays true is already rejected by `_validate_config` with a 400 and never written, and "client_id is required" is the more useful answer. `delete_user` already refused to remove the last admin, so that third way in was closed.

## 4. Publish the four PEMs as one unit

`_atomic_binary_copy` is atomic per file. There was no atomicity across the four, the exception handlers roll nothing back, and `privkey.pem` is **last** in `CERTIFICATE_FILES`. A failure on the fourth file left a new `cert.pem`/`chain.pem`/`fullchain.pem` beside the previous `privkey.pem` — the pair `/download` serves and deploy hooks ship, and it cannot complete a handshake.

The permanence is what made it dangerous: the next renewal fingerprints `live/`, finds it fresh, concludes `renewed=False` and returns *before* the copy loop. Nothing retried it; `check_renewals` booked it as `skipped_not_due`. No failure, no event, no email.

`_publish_flat_files` stages all four then promotes them, so a failure leaves the previous generation intact and internally consistent. The no-op branch now reconciles against `live/` and republishes when they disagree, reporting what it repaired — that is what heals an instance already wedged.

## Verification

- full suite green: 2853 passed, 6 skipped
- every fix disabled in turn to confirm its test goes red; the pre-fix four-file copy reproduced directly (leaves `cert.pem` NEW, `privkey.pem` OLD)
- `flake8` gate clean; `bandit` severities identical to `main` (29 LOW both sides)
- no new dependencies

Not included, deliberately: `/api/activity` is readable by the viewer role while the audit-log endpoints are admin-only. That looks like a product decision rather than a defect — the activity page is built for the viewer role — so it wants a call from @fabriziosalmi rather than a silent change.

### DIFF
--- a/modules/api/resources.py
+++ b/modules/api/resources.py
@@ -3036,7 +3036,15 @@ def create_api_resources(api, models, managers):
                 pre_restore_backup = None
                 if create_backup:
                     current_settings = settings_manager.load_settings()
-                    pre_restore_backup = file_ops.create_unified_backup(current_settings, "pre_restore")
+                    # include_secrets=True: this archive exists for exactly one
+                    # purpose — putting the instance back if the restore below
+                    # goes wrong. A masked rollback cannot recover a single
+                    # credential, which makes it a file that looks like a
+                    # safety net and is not one. It never leaves the host and
+                    # is written chmod 0600, and the opt-in is audit-logged
+                    # with the restore entry below.
+                    pre_restore_backup = file_ops.create_unified_backup(
+                        current_settings, "pre_restore", include_secrets=True)
                     logger.info(f"Created pre-restore backup: {pre_restore_backup}")
 
                 # Restore from unified backup
--- a/modules/core/auth.py
+++ b/modules/core/auth.py
@@ -810,7 +810,16 @@ class AuthManager:
                     settings['users'] = stored_users
 
                 try:
-                    self.settings_manager.update(_touch, "user_management")
+                    # No backup for a plain login. save_settings writes a full
+                    # unified ZIP of settings + certificates for every reason
+                    # that is not None and then prunes to MAX_BACKUPS_PER_TYPE
+                    # (50), so recording last_login evicted a real restore
+                    # point every time somebody signed in. A hash upgrade IS
+                    # worth a restore point; a timestamp is not. The
+                    # backup_reason=None idiom already exists for exactly this
+                    # (api_keys last_used_at).
+                    self.settings_manager.update(
+                        _touch, "password_hash_upgraded" if upgraded else None)
                     if upgraded:
                         logger.info(
                             f"Upgraded the stored password hash for '{username}'"
--- a/modules/core/auth.py
+++ b/modules/core/auth.py
@@ -1004,11 +1004,50 @@ class AuthManager:
         OIDC/SSO-only deployment — but the instance stayed world-open because
         local auth was never enabled. A fresh install with no operator-provided
         credential is unchanged: it stays in setup mode so onboarding works."""
+        return self.setup_mode_for(self.settings_manager.load_settings())
+
+    def setup_mode_for(self, settings):
+        """``is_setup_mode()`` evaluated against an arbitrary settings dict.
+
+        One definition, two callers: the live predicate above, and
+        ``would_open_setup_mode`` below, which asks the same question about a
+        change that has not been written yet. Keeping them as one function is
+        the point — a guard that re-implements the predicate it defends drifts
+        away from it, and the drift is invisible until the day it matters.
+        """
         if self.has_operator_bearer_token():
             return False
-        if self._is_oidc_configured():
+        oidc = settings.get('oidc') or {}
+        if oidc.get('enabled') and oidc.get('issuer_url') and oidc.get('client_id'):
             return False
-        return not (self.is_local_auth_enabled() and self.has_any_users())
+        users = settings.get('users') or {}
+        return not (settings.get('local_auth_enabled', False) and len(users) > 0)
+
+    def would_open_setup_mode(self, candidate_settings):
+        """True iff applying *candidate_settings* would leave this instance
+        serving every gated endpoint to anonymous callers as admin.
+
+        Setup mode is not a mild state. ``_authenticate_request`` returns
+        ``{'username': 'setup_user', 'role': 'admin'}`` for a caller with no
+        credential at all, which is enough to read /api/settings, mint API
+        keys, create admin users and download private keys.
+
+        It exists so a fresh install can be bootstrapped, and it closes as soon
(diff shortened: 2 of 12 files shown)


fabriziosalmi:
### CodeQL: 11 new alerts, and why they are not new risk

The check reports 9 high `py/path-injection` and 2 medium `py/log-injection`, all in `modules/core/certificates.py`, all inside the two methods this PR adds. I checked them rather than waving them through, and the conclusion is that the sinks moved, they were not created.

**The same sinks existed before.** The old renewal loop did `src_file = src_dir / file_name` / `dest_file = dest_dir / file_name` and copied between them (`origin/main:certificates.py:1874-1875`). This PR moves that into `_publish_flat_files` and `_stale_flat_files`. Two named methods with explicit `read_bytes`/`write_bytes`/`copymode`/`replace` calls expose more distinct sink expressions to CodeQL than one inline loop did, which is why nine appear where two were tracked.

**Both rule IDs are already settled policy in this repo.** On `main` there are 27 dismissed `py/path-injection` alerts — 17 in `utils.py`, 7 in `storage_backends.py`, **2 already in `certificates.py`** — dismissed as false positive with this reasoning: *"`_validate_domain_path` (resources.py:48) rejects `..` `/` `\` NUL, enforces `_DOMAIN_RE`, then resolves and checks containment under the cert dir before returning. CodeQL misses that through the helper's return."* That is exactly the flow here: `domain` reaches `renew_certificate` from routes that ran that validator, or from `settings['domains']`, which only those routes write.

`py/log-injection` is the repo's documented won't-fix: default logging is JSON and `JSONFormatter` escapes CRLF at the sink. I deliberately did **not** scrub the domain in only my two new `logger.warning` calls — doing that in 2 places out of ~87 would make the codebase inconsistent about a decision that was taken globally.

**What I have not done:** dismissed these nine alerts. Marking security alerts as false positives is a call for @fabriziosalmi, not something to do quietly inside a PR. If the policy stands, they want the same dismissal comment as the existing 27.

### `acme-dns-live`

Failed with `acme-dns never became ready` — the container never came up, before any application code ran. That job passed on the last three `main` runs, so I have re-run it rather than assume either way; if it goes green it was a flake in the container start, and if it fails again it needs a real look. Not a required check either way.

Issue #666, opened 2026-09-06
create_certificate is F(115), ~590 lines, 24 parameters — decompose it


fabriziosalmi:
**Evidence.** `modules/core/certificates.py:1257`, cyclomatic complexity **115**, ~590 lines, a 24-parameter signature, nesting 5-7 deep. `renew_certificate` (`:1859`) is F(62) and ~404 lines. `CertificateManager` has 57 methods and is simultaneously the domain model, the certbot adapter, the storage orchestrator, the cache and the event publisher. `radon mi` scores the file at the clamped floor of 0.00.

**Impact.** A function with 115 independent paths cannot be branch-tested to any meaningful degree. Every fix mutates a 600-line hot path, and create/renew are near-duplicate logic that has repeatedly drifted — the inline comments in the file memorialise the resulting regressions (privkey copied last, metadata clobbered, renewal omitting the CA bundle). This is the highest-risk single unit in the system, and it is the one that handles private keys.

**Definition of done.** The issuance path is composed of units that can be tested and reasoned about independently, and create/renew stop being two copies that drift.

fabriziosalmi:
**Re-measured at `b07f174`** (this issue's own numbers, taken again):

| | when this was written | now |
|---|---|---|
| `create_certificate` complexity | **115** | **18** |
| `create_certificate` length | ~590 lines | 317 lines |
| `create_certificate` parameters | 24 | 18 |
| `renew_certificate` | F(62), ~404 lines | **35**, 380 lines |

The definition of done had two halves. The first — *"the issuance path is composed of units that can be tested and reasoned about independently"* — is met for the create path: 18 independent paths is a unit that can be branch-tested, against 115 which could not. The second — *"create/renew stop being two copies that drift"* — is met where it was costing the most: `_prepare_issuance`, `_build_issuance_command`, `_IssuanceArtifacts` and `_store_in_backend` are now single units both paths go through, and the comment at `certificates.py:1072` records the specific drift (#423) that made it necessary.

**What has not moved, and is worth keeping honest about:**

- `renew_certificate` is still **35**, the highest in the file after `get_certificate_info` (25) and `_build_issuance_command` (23). The complexity gate is at 115, so none of these is policed by anything today.
- `CertificateManager` now has **84 methods**, up from the 57 this issue recorded. The god-object half of the Evidence section got worse while the function-level half got better. That part overlaps #668 (the layering inversion) more than it belongs here.

[...]

Issue #590, opened 2026-08-21
constants.py: six defaults with no reader, one of them contradicting the code (session timeout has three values)


fabriziosalmi:
Found during the 2026-08-21 cross-review; not for v2.26.0 (no operator-facing defect), but the same shape as three things fixed that day — an interface that says something the code does not do.

**Six of the nine names in `modules/core/constants.py` have no reader outside the file** (verified: no `import *`, no `getattr(constants, ...)`): `_FILESYSTEM_ARTIFACT_DIR_NAMES`, `DEFAULT_RENEWAL_THRESHOLD_DAYS = 30`, `DEFAULT_LOGIN_RATE_LIMIT = 5`, `DEFAULT_LOGIN_RATE_WINDOW`, `DEFAULT_SESSION_TIMEOUT_HOURS = 24`, `DEFAULT_CACHE_TTL = 300`.

**The real values live elsewhere, and one contradicts it:**
- `constants.py: DEFAULT_SESSION_TIMEOUT_HOURS = 24` — `auth.py:75` reads `SESSION_TIMEOUT_HOURS` with default **8**; and `auth_routes.py:73` sets the session cookie with `max_age=8 * 60 * 60` regardless of either. Three versions of one number; the env var is read but the cookie ignores it (audit finding auth_routes.py:73).
- `constants.py: DEFAULT_RENEWAL_THRESHOLD_DAYS = 30` — duplicated by hand as `'renewal_threshold_days': 30` in two places of settings.py.

**Decide, then make it true:** either `constants.py` is the source of defaults — then the callers read it (auth.py, settings.py, the cookie `max_age`), and `SESSION_TIMEOUT_HOURS` actually drives the cookie — or it is not, and the six lines go, because a constants file that lies is worse than one that does not exist.

PR #734, opened 2026-09-07
refactor(#666): pin the certbot command, then extract _prepare_issuance (step 1)

fabriziosalmi:
First step of #666. **`create_certificate` F(115) → F(65).** Deliberately not the whole decomposition — see *What is left* below.

## The net, first

`create_certificate` is the highest-risk single unit in the system and the one that handles private keys. The inline comments in the file memorialise what past changes cost: privkey copied last, metadata clobbered, renewal omitting the CA bundle.

So before anything moved, `tests/test_issuance_command_is_a_contract.py` pins the thing a decomposition must not change: **for eleven inputs covering the branches inside that 517-line try, the exact certbot argv.**

Each case exercises a different branch, and the diffs show it: `http-01` replaces the whole DNS authenticator block, `replace` **drops** `--key-type`/`--elliptic-curve` (deliberate — a reissue keeps the lineage's key unless explicitly re-keyed), `domain alias` switches to `--manual` with two hooks, a provider with no configured propagation falls back to 120s.

Two controls on the matrix itself, not just the cases:
- every case must build a command that **differs from the baseline** — a matrix whose entries all agree looks thorough while pinning one branch;
- the union of the cases must contain the flags a decomposition could silently drop (`--webroot`, `--staging`, `--force-renewal`, `--manual-auth-hook`, …).

Expectations are **inline, not a golden file with a regenerate flag**. A changed command then shows up as a diff a reviewer reads, instead of a file someone re-blesses.

Confirmed it bites: removing `--non-interactive` or `--cert-name` from the builder turns eleven cases red.

## The extraction

The first third of the try block was all resolution — which CA, which challenge type, which DNS provider and account, which key shape, which domains, whether the domain is already issued. It moves out verbatim behind a frozen `_PreparedIssuance` record.

A record rather than fourteen locals threaded down a 590-line function: what crosses between phases is now visible and cannot be reassigned halfway through.

**Named `_prepare_issuance`, not `_resolve_…`** — it is not pure. It creates the per-domain output directory and, for HTTP-01, the webroot challenge directory. Both were side effects of this stretch before the move; a name claiming purity would be the same class of defect as a constants file that lies (#590).

**The net earned its keep immediately:** `cert_dir` crosses the boundary and I did not carry it. Eleven cases went red until it was.

## What is left, and why it is not here

| | |
|---|---|
| build the command (CA/EAB, replace flags, DNS credentials, alias hook) | ~160 lines |
| run + collect artifacts | |
| persist metadata + storage | |
| `renew_certificate` F(62), and its near-duplication with create | the actual "stop being two copies that drift" half |

The command-builder step has a real hazard that deserves its own reviewable change rather than being buried here: `credentials_file`, `extra_credential_files` and `ca_extra_env` are created during command building and removed by the `finally`. A naive extraction that returns them loses cleanup when the builder raises **partway** — a live cloud private key then outlives the operation on disk. The fix is to pass a mutable artifacts record the builder fills as it goes, which also makes that guarantee explicit for the first time instead of implicit in three locals.

## Verification

- Unit suite 3595 → 3608 (+13, the net), all green
- **Real-cert E2E: 13 tests, 230s**, LE staging via Cloudflare — this touches the issuance pipeline, so it was run before opening the PR rather than left to the release gate
- gated flake8 clean, per-module coverage floors met

### DIFF
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -108,6 +108,30 @@ _KNOWN_METADATA_KEYS = _REISSUE_OWNED_METADATA_KEYS | frozenset({
 })
 
 
+def _reject_path_escaping_domain(domain):
+    """Refuse a domain that could escape ``cert_dir`` when used as a path.
+
+    ``domain`` becomes a directory name under the certificate root, the certbot
+    ``--config-dir``, and the key of the per-domain lock. A URL-form value whose
+    netloc passed ``validate_domain`` but was kept raw ("https://x/../y") would
+    escape ``cert_dir`` and drop the ACME account private key under the public
+    ``/.well-known/acme-challenge`` webroot.
+
+    Reject rather than normalise: normalisation is the source's job, and
+    reaching this sink with path characters means something upstream failed, so
+    the request must not proceed.
+
+    Module-level and called from every entry point rather than written inline
+    once (#666). While it lived at the top of ``create_certificate`` it
+    protected the code below it by position; extracting that code into
+    ``_prepare_issuance`` produced a unit that did not enforce its own
+    precondition, which is exactly the shape CodeQL flags — and it was right.
+    """
+    if (not domain or '/' in domain or '\\' in domain
+            or '..' in domain or '\x00' in domain):
+        raise ValueError('Invalid domain name')
+
+
 @dataclass(frozen=True)
 class _PreparedIssuance:
     """Everything ``create_certificate`` resolves before it builds a command.
@@ -1468,6 +1492,11 @@ class CertificateManager:
         Raises the same exceptions the inline code did — FileExistsError,
         ValueError, RuntimeError — so the caller's handlers are unchanged.
         """
+        # create_certificate rejects these before taking the per-domain lock,
+        # and this repeats it rather than trusting the caller: `domain` is used
+        # here to build cert_dir paths, and a unit that assumes its caller
+        # validated is only safe by position.
+        _reject_path_escaping_domain(domain)
         settings = None
         # Settings are loaded lazily and at most once: several branches
         # below need settings (CA default, challenge type, DNS provider,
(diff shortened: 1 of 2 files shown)


fabriziosalmi:
## CodeQL: six alerts, and it was right about the shape

Two `py/path-injection`, four `py/log-injection`, all inside `_prepare_issuance` — code this PR *moved*, not code it wrote. They are absent on `main`.

That is not a reason to dismiss them, and I did not treat it as one. **The extraction genuinely weakened something.** The guard's own comment says it exists to be "the last line that every path must cross"; while it sat at the top of `create_certificate` it protected the code below it **by position**. Moving that code into a method produced a unit that did not enforce its own precondition — safe today because there is one caller, unsafe the moment there are two.

So the check is now `_reject_path_escaping_domain`, module-level, called from **both** entry points (commit 3e2cb05). The invariant belongs to the code that depends on it rather than to where it happens to sit.

## Why the alerts remain, and what I did instead

CodeQL does not model an interprocedural guard that *raises*, so it still cannot see the barrier. The two remaining options were to duplicate the two-line check inline where the analyser can see it, or to dismiss with evidence.

**I rejected the inline duplication.** A security check with two copies is a security check that drifts — that is the same failure mode as #590's constant that said 24 while the code used 8.

Dismissed with evidence instead, having verified by execution that `cert_dir / domain` cannot leave the root for any escape shape:

| input | rejected | escapes root |
|---|---|---|
| `/etc` | yes | — |
| `../../etc` | yes | — |
| `a/../..` | yes | — |
| `..` | yes | — |
| `.` | no | no (resolves to the root itself) |
| `~`, `~root` | no | no (pathlib does not expand) |
| `C:` | no | no (not special on POSIX) |
| embedded NUL | yes | — |

The NUL case is worth a note: it crashed the *probe* — `pathlib.resolve()` raises `ValueError: embedded null character` — which is precisely why the guard has to reject it before it reaches a path operation, and does.

PR #736, opened 2026-09-07
refactor(#666): one propagation formula instead of four, and one had drifted (step 3)

fabriziosalmi:
Step 3 of #666, on top of #735. **`renew_certificate` F(62) → F(56)** — and this is the first step that touches the half the issue actually cares about: *"create/renew stop being two copies that drift."*

## Four copies, not two

The DNS-01 propagation wait was computed in **four** places:

| where | clamped to 1..3600? |
|---|---|
| `create_certificate`, every provider | yes |
| `renew_certificate`, alias branch | yes |
| `renew_certificate`, `custom-script` | yes |
| `renew_certificate`, **acme-dns alias branch** | **no** |

Same settings key, same arithmetic, hand-written each time. One of them carries a comment reading *"Mirror the create path"* — a copy that announces it is a copy.

## They had already drifted

That is the point of the issue rather than a hypothetical risk. Three copies bounded the value to 1 second .. 1 hour. The acme-dns alias branch did not, so:

- `dns_propagation_seconds: 0` asked the hook to validate **before the TXT record existed**;
- `dns_propagation_seconds: 86400` held the per-domain lock for a **day**, which under `check_renewals` (serial, `max_instances=1`) stalls every subsequent renewal behind it.

**Behaviour change, deliberate:** that branch now clamps like the other three. Called out here rather than folded into "refactor", because it is the one thing in this PR that changes what the program does.

## How the other two were found

I found two of the four by reading. The other two were found by the test, which asserts on the source that no call site reads `dns_propagation_seconds` with its own arithmetic again — it failed immediately, twice, on copies I had missed.

That guard needed correcting too: its first version asserted the bare *name* was absent and tripped on a comment that merely mentions it. It matches the **read** now (`.get('dns_propagation_seconds'`), not the word. Confirmed it bites by reintroducing a copy in `renew_certificate`.

## What deliberately did not change

`certbot renew` replays the flags stored in `renewal/<domain>.conf`, so the renew path does not re-emit `--{plugin}-propagation-seconds` the way create does. That asymmetry is real — an operator who raises the setting after issuance keeps the old value until the next create or reissue — but it is **behaviour, not duplication**, and changing it is not a refactor. It is recorded in the test module docstring so the next reader does not "fix" it by accident.

## Verification

17 tests on the helper (bounds, unparseable values, a missing map, `None` settings from the lazy load) plus the source guard. Suite 3615 → 3632 (+17), gated flake8 clean, **real-cert E2E 13 tests / 238s** against LE staging.

## Still to come

`renew_certificate` is still F(56): the DNS credentials setup (`create_config_file` → `extra_credential_files` → `configure_certbot_arguments`) is duplicated between create and renew as well, and renew keeps its own cleanup trio rather than the `_IssuanceArtifacts` record from #735.

### DIFF
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -108,6 +108,34 @@ _KNOWN_METADATA_KEYS = _REISSUE_OWNED_METADATA_KEYS | frozenset({
 })
 
 
+def _propagation_seconds(settings, dns_provider, strategy):
+    """How long to wait for a DNS-01 TXT record to propagate, in seconds.
+
+    One formula with one home (#666). It existed twice — once in
+    ``create_certificate`` for every provider, once in ``renew_certificate``
+    narrowed to ``custom-script`` — with identical arithmetic and a comment on
+    the second saying "Mirror the create path". Two copies that must agree, one
+    of which announces that it is a copy, is the shape this issue is about.
+
+    Falls back to the strategy default when the settings map has no entry, when
+    the entry will not parse, or when settings cannot be read at all, and
+    clamps to 1 second .. 1 hour so a typo cannot make an issuance hang or
+    return before the record is visible.
+    """
+    try:
+        propagation_map = (settings or {}).get('dns_propagation_seconds', {}) or {}
+    except Exception as e:
+        logger.debug("Failed to read dns_propagation_seconds: %s", e)
+        propagation_map = {}
+
+    default_seconds = strategy.default_propagation_seconds
+    try:
+        seconds = int(propagation_map.get(dns_provider, default_seconds))
+    except (ValueError, TypeError):
+        seconds = default_seconds
+    return max(1, min(3600, seconds))
+
+
 def _reject_path_escaping_domain(domain):
     """Refuse a domain that could escape ``cert_dir`` when used as a path.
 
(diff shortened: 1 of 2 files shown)


fabriziosalmi:
**Correction: there was no behaviour change. I got this wrong.**

This PR claimed the acme-dns alias branch never clamped the propagation value, and called unifying it a deliberate behaviour change. That is not what the code did.

All four copies clamped to 1..3600. Three clamped where the value was computed; the acme-dns alias branch clamped **at the call site**, four lines further down:

```python
renew_propagation = int(propagation_map.get(dns_provider, strategy.default_propagation_seconds))
...
alias_hook_config = self._create_dns_alias_hook_config(
    dns_provider, dns_config, acme_dns_alias,
    max(1, min(3600, renew_propagation)),   # <- it was here
)
```

I read the assignment, saw no clamp, and did not read far enough. So:

- **`dns_propagation_seconds: 0` never asked the hook to validate early**, and `86400` never held the per-domain lock for a day. Those consequences did not exist.
- The unification changes **no behaviour at all**. What it removes is four places where a future edit could move one clamp and not the others.

Nothing shipped differently — the merged code clamps exactly as before, and the E2E and suite were green because the behaviour was identical either way. What was wrong was the claim, in the commit message, this PR body, and a code comment.

Fixed in the follow-up: the misleading comment is gone (along with the now-redundant second clamp it left behind), and the test module docstring says what actually happened.

The rest of the PR stands: four copies of one formula, one of them announcing itself as a copy, and the source-level guard finding two of them that I had missed by reading.

Issue #672, opened 2026-09-06
Route handlers reach into private manager methods; CertificateService is half-adopted


fabriziosalmi:
**Evidence.** `CertificateDetail.patch` and `delete` in `modules/api/resources.py` call `certificate_manager._load_metadata` and `_save_metadata` — private methods — directly, and perform the metadata read-modify-write inline. Web routes do function-body imports of core internals (`modules/web/misc_routes.py:379,401,404,467`). `modules/web/cert_routes.py` re-implements domain sanitisation and path handling in parallel with `resources.py`.

**Impact.** `CertificateService` was created explicitly to end this duplication and is only half-adopted: create/renew/reissue go through it, config-edit and delete bypass it. The manager cannot change how it stores metadata without breaking routes, and security-relevant path-traversal validation exists in two places — which is precisely how one copy gets fixed and the other does not.

**Definition of done.** The HTTP layers are thin adapters over one service boundary, with no reach-through to private internals and no duplicated path validation.

fabriziosalmi:
Concrete instance for this issue, from #640 / #726.

Event publishing is a caller responsibility rather than the service's. `CertificateService.create()` issues the certificate but publishes nothing; each call site is expected to publish `certificate_created` afterwards. Five call sites do. The adoption handler did not — so an adopted certificate ran no deploy hook and left a stale deployment-status verdict cached, with no notification. That is the same shape this issue is about: a service that owns an operation but not its full contract, with the remainder duplicated across handlers that can each forget a piece.

@ratiugtun proposed the root-cause fix in the original report: publish from inside `CertificateService.create()` so every caller gets it for free. It was deliberately not done in #726 because the four call sites that already publish explicitly would double-publish, and untangling that is this issue's work rather than a bug fix's.

So when `CertificateService` is finished, the event publish should move inside it and come out of the handlers — with the caveat that the payloads are not identical today (the adopt path carries `adopted` and `fingerprint`; the create path carries `san_domains`, `dns_provider`, `ca_provider`), so the service needs a way to accept per-caller context rather than a single fixed shape.

fabriziosalmi:
Both halves done.

**Duplicated path validation** (#742): the rule the issue said existed "in two places" existed in **six** — the API layer, the web layer, the storage backends, the issuance guard, client certificate identifiers, and backup filenames. I counted two by reading; the source guard written for the consolidation found the third and fourth on its first run and the fifth on its second.

[...]

Issue #668, opened 2026-09-06
Layering inversion: the composition root lives in modules/core and imports api and web


fabriziosalmi:
**Evidence.** `modules/core/factory.py:30-32` imports from `modules.api` and `modules.web`; both of those import from `modules.core`. The resulting package-level cycle is avoided only because the offending imports are deferred to the bottom of the file.

**Impact.** "Core" is not an innermost domain layer — it is entangled with the web framework, so the domain cannot be reasoned about, reused or tested independently of Flask. The dependency direction of the whole codebase is inverted at its root.

**Definition of done.** The dependency direction points inward and the composition root sits above the layers it composes.

Issue #667, opened 2026-09-06
create_api_resources is a 3,708-line function holding 42 Resource classes


fabriziosalmi:
**Evidence.** `modules/api/resources.py:333` does not return until line 4040 — 92% of the file is one function body, containing 42 Flask-RESTX `Resource` classes and 60 nested defs as closures over `managers`. Individual endpoint closures are themselves large (269 lines at `:3445` with nesting depth 9; 250 at `:2017`; 243 at `:521`). The file also mixes raw TLS/SMTP socket probing into the HTTP-routing module.

**Impact.** Dependency injection by lexical closure means no route class can be imported, instantiated or unit-tested without constructing the entire manager graph — which is why this layer is the least covered (see the coverage issue in phase-3). It is also beyond what any tool or reviewer can hold in view.

**Definition of done.** Route classes are independently importable and testable, and HTTP routing does not contain socket-probing logic.

fabriziosalmi:
### Progress: half the file, and a map of what is left

Merged so far: context+contract (#691), cache (#692), health (#693), backup (#694).
In flight: storage (#695), inventory (#696), settings/DNS (#697), CA (#698).

| | at #669 open | now |
|---|---|---|
| `create_api_resources` complexity | 559 | **249** |
| `resources.py` | 4040 lines | **2023** |

### What is left, measured rather than guessed

The remaining 16 classes are 1507 lines. Asking the AST which names each captures
gives the grouping directly:

| group | classes | lines |
|---|---|---|
| downloads | DownloadCertificate, DownloadCertificateFile | 344 |
| lifecycle/jobs | CreateCertificate, RenewCertificate, CertificateReissue, CertificateJob, CertificateJobs, CertificateAutoRenew | 391 |
| core | CertificateList, CertificateDetail | 374 |
| deployment | CertificateDeploymentStatus, CertificateDeploymentBrowserReports, CertificateRunDeploy | 280 |
| discovery/DNS | ZombieScan, CheckDNSAlias, CertificateDNSAliasCheck | 118 |

### The one blocker, and why the downloads group was not done next

Every group so far needed only the context and the thin helper wrappers. These do
not: **`_validate_domain_path` is called from 13 sites still inside the closure**, so
a module that uses it would have to import `resources.py`, which imports it back.

That helper needs a shared home first — with `_DOMAIN_RE`, probably as a small
`modules/api/path_validation.py`, re-exported from `resources.py` so the existing
[...]

PR #742, opened 2026-09-07
refactor(#672): one home for domain-to-path validation

fabriziosalmi:
Half of #672 — the duplicated path validation. (The `CertificateService` reach-through is the other half and is not in this PR.)

## Six copies, not two

The issue says the rule "exists in two places". It existed in **six**:

| | |
|---|---|
| `modules/api/path_validation.py` | the API layer |
| `modules/web/routes.py` `_sanitize_domain` | the web layer |
| `modules/core/storage_backends.py` | secret keys |
| `modules/core/certificates.py` | the issuance guard |
| `modules/api/client_certificates.py` | client identifiers |
| `modules/api/resources_backup.py` | backup filenames |

Same forbidden characters in all six, three different return conventions, and two of them whole functions differing only in the name of their regex.

`modules/core/domain_paths.py` is now the only definition — the character rule plus the three patterns (certificate domain, storage domain, client identifier), consistently anchored. It lives in `core` rather than `api` because `web` and the storage backends need it too, and having core import from api would deepen the layering inversion in #668.

## Consolidating closed a live inconsistency

Some of these copies predated a correction the others had already received, so **the same input was accepted at one entry point and refused at another**, depending only on which route it arrived through. That is the concrete cost of duplicated validation, and it was present here rather than hypothetical.

## How three of the six were found

I counted **two** by reading. The source guard I wrote for the consolidation found the third and fourth on its first run, and the fifth on its second. The count went from two to six inside an hour.

That is the actual argument for the guard, and it is the reason the tests assert **every entry point reaching the same verdict for the same input** rather than only exercising the shared helper — the latter would pass just as well with five stale copies still in the tree.

`resources_backup` was the one copy already correct, and instructively so: it rejects every control character, with a comment explaining a problem fixed in an earlier session. Whoever wrote it understood the issue; the other copies never knew.

## Verification

| mutation | killed |
|---|---|
| the domain pattern's end-anchor reverts | ✅ |
| the identifier pattern's end-anchor reverts | ✅ |
| NUL leaves the forbidden set | ✅ |
| `web/routes` builds the path itself again | ✅ (8) |

Suite 3676 → 3710 (+34), gated flake8 clean, **real-cert E2E 13 tests / 238s** against LE staging.

### DIFF
--- a/modules/api/client_certificates.py
+++ b/modules/api/client_certificates.py
@@ -1,23 +1,21 @@
 import logging
-import re
 from flask import request, send_file, Response
+from ..core.domain_paths import IDENTIFIER_RE, is_path_safe_segment
 from werkzeug.exceptions import HTTPException
 from flask_restx import Resource, fields, abort
 from io import BytesIO
 
 logger = logging.getLogger(__name__)
 
-# Validation pattern for client certificate identifiers
-_SAFE_IDENTIFIER_RE = re.compile(r'^[a-zA-Z0-9][a-zA-Z0-9._-]{0,127}$')
-
-
 def _validate_identifier(identifier):
-    """Validate client certificate identifier to prevent path traversal."""
-    if not identifier or '..' in identifier or '/' in identifier or '\\' in identifier or '\x00' in identifier:
-        return False
-    if not _SAFE_IDENTIFIER_RE.match(identifier):
-        return False
-    return True
+    """Validate client certificate identifier to prevent path traversal.
+
+    The character rule and the pattern both live in ``core.domain_paths`` (#672).
+    They were written here as a fourth copy, and carried the same `$` anchor the
+    other copies had.
+    """
+    return (is_path_safe_segment(identifier)
+            and bool(IDENTIFIER_RE.match(identifier)))
 
 
 def create_client_certificate_models(api):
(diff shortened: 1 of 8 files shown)


fabriziosalmi:
## CodeQL: one alert, on the containment check itself

`py/path-injection` on `resolved = cert_dir.resolve()` in `domain_paths.validate_domain_path`.

That line **is** the safety check. The function's whole job is to build the candidate path and then confirm it resolves under the base directory, returning `(None, error)` when it does not — and it only gets there after `is_path_safe_segment` and `DOMAIN_RE` have both passed, in the same function, three lines above.

It appears now for the same reason as the alerts on #734, #735 and #739: the code **moved**, so it is new in this diff. The identical lines were not flagged in `api/path_validation.py`, where they have lived since #667. CodeQL does not model `is_path_safe_segment` as a barrier because it is an interprocedural predicate.

Dismissed with evidence rather than worked around. The evidence is the test this PR adds: **every one of the six entry points refuses the same set of escapes** — `..`, `../etc`, `a/b`, `a\b`, `/etc/passwd`, NUL, empty — asserted per entry point so a single stale copy fails by name.

Inlining the character rule back into this function to satisfy the analyser would undo exactly what the PR is for.

PR #747, opened 2026-09-07
fix: constrain the extras install layer, and record the hash-pinning decision (#686)

### LINKED ISSUE #686: Hash-pin the dependency sets (--require-hashes), or decide deliberately not to
Split out of #659, which also covered SBOM/provenance (done, #684) and build-time OS upgrades (done, #685). This is the remaining third, deliberately separated because it is a project rather than a cleanup.

## What is missing
Version pins (`==`) do not defend against a **re-published or compromised artifact at that same version**. Hash pinning does. Today there is no `--require-hashes`, no lockfile carrying hashes, and nothing verifying that the wheels installed are the wheels that were reviewed.

## Why it is not a small change
`--require-hashes` is all-or-nothing per install: **every** requirement, including every transitive dependency, must carry a hash, or pip refuses the whole file. That collides with how this image is actually built:

- **12 requirements files.** Each would need a fully-resolved, hash-locked companion.
- **Layered installs.** The Dockerfile installs `REQUIREMENTS_FILE` and then `EXTRA_REQUIREMENTS` as a second layer, and documents real combinations (`requirements-azure.txt requirements-storage-all.txt`, `requirements-aws.txt requirements-gcp.txt`). Under `--require-hashes` each layer must be independently complete, so the combinations multiply rather than compose.
- **Multi-arch.** amd64 and arm64 resolve different wheels for packages with binary distributions, so a single hash list does not cover both platforms.
- **~192 installed packages** in the full image.

## Why doing it badly is worse than not doing it
A half-hashed set fails installs with errors that name the wrong cause, and the reliable human response is to route around the mechanism — `--no-deps`, a second unhashed install, or dropping the flag entirely. The project then carries the maintenance cost of lockfiles **and** the false confidence of a guard that people bypass.

## What partially covers the gap today
- Every direct dependency is version-pinned, and phase-2 closed the gaps where a set could resolve something the stack rejects (#657, #658).
- The published image now ships an **SPDX SBOM and SLSA provenance** (#684), so what actually shipped is auditable after the fact, and the base-image digest is recorded.
- The base image is digest-pinned and OS changes now arrive as reviewable digest bumps (#685).

That is auditability, not integrity — it tells you what you got, not that it is what was intended. The gap is genuine; the point is that it is the *narrower* remaining risk, not the urgent one.

## What a real answer needs
1. A generator (pip-tools / uv) producing hash-locked files per variant, per platform, in CI rather than by hand.
2. A decision on whether the layered `EXTRA_REQUIREMENTS` model survives, or whether variants become fully-resolved standalone sets.
3. A refresh cadence, since a hash lockfile that nobody regenerates becomes a reason not to patch.
4. A guard that the lockfiles match the requirements files, or they drift — the failure mode this milestone keeps finding.

**Recommendation: do not attempt this incrementally.** Either commit to the generated-lockfile design above, or record explicitly that version pinning plus SBOM is the accepted posture, so the absence is a decision rather than an oversight.

fabriziosalmi:
Closes #686.

The issue asked for `--require-hashes` **or** an explicit decision not to. The
answer is **not now, recorded as a decision** — and the reachable half of the
gap it was standing in for is closed here.

### The reachable half

Each `EXTRA_REQUIREMENTS` file is a separate pip resolution that knows nothing
about what the base layer pinned, so it is free to move those pins. Measured,
not reasoned about — with the base stack installed:

```
$ pip install --dry-run "cryptography<46"      # as a second layer
  cryptography 46.0.7 -> 45.0.7
```

45.0.7 is below the floor `pyopenssl==26.0.0` requires, so that build ships an
interpreter where `certbot --version` no longer answers. With the constraint:

```
$ pip install --dry-run -c requirements.txt "cryptography<46"
  ERROR: Cannot install cryptography<46 because these package versions have
  conflicting dependencies.
```

**This has already happened here once.** CI's own comment records it:
`requirements-aws.txt` pinned a plugin above the certbot pin and pip upgraded
certbot 2.10.0 → 3.3.0 *in a build that reported success*. That was fixed by
correcting the pin; the mechanism that allowed it stayed open. The storage
extras walk straight back into it — they carry unbounded `>=` requirements on
`azure-identity`, `boto3` and `azure-keyvault-*`, any of which can ask for a
newer `cryptography` in a future release.

Every documented combination was resolved under the constraint before changing
anything:

| base | extras | |
| --- | --- | --- |
| `requirements.txt` | `requirements-azure.txt requirements-storage-all.txt` | ok |
| `requirements.txt` | `requirements-aws.txt requirements-gcp.txt` | ok |
| `requirements-minimal.txt` | `requirements-azure.txt requirements-storage-all.txt` | ok |

The trade is real and worth stating: a variant build that would previously have
shipped a broken stack now fails loudly at build time.

### Two advertised combinations were never resolved

The CI step's comment claims it resolves *"every file and every combination the
Dockerfile and docker-compose.yml advertise"*. Both advertise
`requirements-storage-all.txt`, alone and alongside `requirements-azure.txt`,
and neither was in the list. Nothing connected the two places, so nothing said
so. Both are added (checked: both resolve, certbot stays at 2.10.0), and
`test_every_advertised_combination_is_resolved_in_ci` now ties them together.

That test accepts a CI entry only when it is the advertised combination plus one
base file — **not** merely a superset. Resolving azure+storage-all together
proves storage-all is *installable*, since a subset of requirements is easier to
satisfy; it does not prove the narrower resolution someone actually gets is the
one that was checked.

### The decision, on record

`SECURITY.md` gains a **Supply-chain posture for Python dependencies** section:
what is guaranteed (exact pins, cross-file pin agreement, digest-pinned base,
SBOM + provenance, and now a constrained extras layer), what is **not** (no hash
verification — a re-published artifact at the same version is undetected), why
hash pinning is not attempted incrementally (all-or-nothing per install × 12
files × layered variants × 2 architectures × 121 resolved packages, and a
half-hashed set gets routed around with `--no-deps`), and what would change the
answer.

The honest summary, written into the file: **auditability, not integrity.**

### Checks

Two new tests, both mutated to confirm they bite — reverting the Dockerfile to
an unconstrained extras install, and removing either newly-added CI combination.
A control confirms the drift guard stays quiet for CI entries that are extra
rather than advertised. Full unit suite 3774 passed / 31 skipped, gated flake8
clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,12 +252,21 @@ jobs:
       if: matrix.python-version == '3.12'
       run: |
         set -euo pipefail
+        # Kept in step with what the Dockerfile, docker-compose.yml and the
+        # README actually advertise —
+        # tests/test_requirements_are_consistent.py fails if one of them names
+        # a combination this list does not resolve. The storage entries below
+        # arrived that way: the comment above claimed "every combination the
+        # Dockerfile advertises", and two of the three it advertises were not
+        # here (#686).
         combos=(
           "requirements-minimal.txt requirements-extended.txt"
           "requirements.txt requirements-aws.txt"
           "requirements.txt requirements-gcp.txt"
           "requirements.txt requirements-azure.txt"
           "requirements.txt requirements-aws.txt requirements-gcp.txt"
+          "requirements.txt requirements-storage-all.txt"
+          "requirements.txt requirements-azure.txt requirements-storage-all.txt"
         )
         pinned=$(grep -oE '^certbot==[0-9.]+' requirements.txt | cut -d= -f3)
         echo "certbot is pinned at $pinned"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -306,6 +306,45 @@ jobs:
           echo "::endgroup::"
         done
 
+    # The images are published for amd64 AND arm64, and every check above runs
+    # on exactly one of them. That is how `infisical-python>=2.3.6` sat in two
+    # advertised requirements files: 2.3.6 ships an aarch64 wheel, macOS wheels
+    # and Windows wheels, but no manylinux x86_64 wheel and no sdist. It
+    # installs on the arm64 image and cannot be installed on the amd64 one at
+    # all. A resolver running on either architecture alone sees nothing wrong.
+    #
+    # pip can resolve FOR another platform without emulating it. The check is
+    # not "does it resolve" — with --only-binary a source-only distribution
+    # fails everywhere, and most certbot plugins are source-only — but "do the
+    # two architectures AGREE". A file that fails on both is inconclusive, not
+    # broken, and is reported as such rather than counted as covered.
+    - name: Resolve the optional requirements for both published architectures
+      if: matrix.python-version == '3.12'
+      run: |
+        set -euo pipefail
+        skew=0
+        for req in requirements*.txt; do
+          case "$req" in requirements-test.txt) continue;; esac
+          verdicts=""
+          for plat in manylinux_2_28_x86_64 manylinux_2_28_aarch64; do
+            if python -m pip install --dry-run --quiet --no-cache-dir \
+                 --ignore-installed --only-binary=:all: \
+                 --platform "$plat" --python-version 3.12 \
+                 --target "$(mktemp -d)" -r "$req" >/dev/null 2>&1; then
+              verdicts="$verdicts $plat=ok"
+            else
+              verdicts="$verdicts $plat=no"
+            fi
+          done
+          case "$verdicts" in
+            *x86_64=ok*aarch64=ok*) echo "$req: both architectures resolve";;
+            *x86_64=no*aarch64=no*) echo "$req: inconclusive (source-only distributions; not covered by this check)";;
+            *) echo "::error::$req resolves on one published architecture and not the other:$verdicts. An image built for the other architecture cannot install it."
+               skew=1;;
+          esac
+        done
+        [ "$skew" -eq 0 ]
+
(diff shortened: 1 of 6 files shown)


fabriziosalmi:
The CI combination this PR added went red on its first run, and what it found is worth more than the change that added it.

`requirements-storage-all.txt` and `requirements-infisical-storage.txt` both carried `infisical-python>=2.3.6`. That version ships an aarch64 manylinux wheel, macOS wheels and Windows wheels — **and no manylinux x86_64 wheel and no sdist**. So:

* on the **arm64** image it installs fine;
* on the **amd64** image it cannot be installed at all.

Both files are advertised as `EXTRA_REQUIREMENTS` by the Dockerfile and `docker-compose.yml`.

Pinned to `2.3.5`, the last release with complete platform coverage. Measured, not assumed: it is the only compiled dependency in the storage extras — every other one ships a pure-Python `any` wheel.

### Why nothing saw it, and what now does

Every dependency check in this repo runs on one architecture; the images are published for two. A resolver on either alone sees nothing wrong. **My own local verification of this PR reported the combination as fine** — on arm64 macOS, where 2.3.6 exists. The check that caught it was the one running on GitHub's x86_64 runner.

pip can resolve *for* another platform without emulating it, so CI now resolves each optional requirements file for both published architectures and fails when the two **disagree**. Deliberately not "does it resolve": with `--only-binary` a source-only distribution fails everywhere, and most certbot plugins are source-only. A file failing on both is reported as *inconclusive*, not counted as covered — 8 of 11 files get real coverage and the step says which.

`test_the_architecture_check_covers_every_published_architecture` ties the two architectures CI checks to the two the release publishes, so adding a platform cannot silently leave it unchecked.

PR #748, opened 2026-09-07
feat: deploy hooks can be held for a maintenance window (#632)

### LINKED ISSUE #632: [Feature] Scheduled deployment for hooks
We have a need to deploy to multiple locations but at a specific time. Right now it will run the renewal when it's ready, but we don't know exactly when. This would mean that the deployment-script runs at a random time.
For example certificate changes on RDS-server deployment (I know...) this will disrupt end user connectivity.

Schedule Enable Deploy Hooks at a specific time, or have schedules for renewals.

Maybe there is an easy answer to this and I am overthinking it. One option would be ansible and API calls during a maintenance window. But that adds complexity. If anyone has any other ideas let me know :)

fabriziosalmi:
Closes #632.

> We have a need to deploy to multiple locations but at a specific time. Right now it will run the renewal when it's ready, but we don't know exactly when... certificate changes on RDS-server deployment, this will disrupt end user connectivity.

The diagnosis is right. A certificate renews when it is **due** — the sweep fires at 02:00 with an hour of jitter, and only for whatever is inside the threshold that night — and the deploy hook then ran immediately. For a hook that restarts a database or reloads a load balancer, that is an outage at an hour nobody chose.

A `window` separates the two. The certificate still renews whenever it is due, which is driven by expiry and not negotiable. The **deploy** is held until the window next opens.

```jsonc
"window": {
  "start": "02:00",          // required, 24-hour HH:MM
  "end": "04:00",            // required, exclusive
  "days": ["sat", "sun"],    // optional; omit for every day
  "timezone": "Europe/Rome"  // optional IANA name, defaults to UTC
}
```

Absent means **immediately** — what every existing hook does, and keeps doing. The same field works on typed deploy targets (#475), which is usually where it is wanted: a Kubernetes secret rollout is exactly that kind of deploy.

### The calendar lives on its own

`modules/core/deploy_window.py` is pure functions over an explicit `now`, because every interesting case here is a calendar case and none of them is testable through a queue and a scheduler:

* a window may **cross midnight**, and belongs to the day it **starts** on — a `22:00`-`04:00` Friday window is open at 02:00 Saturday. The other reading would silently require Saturday to be ticked as well;
* `end` is **exclusive**, so `02:00`-`04:00` and `04:00`-`06:00` are adjacent rather than overlapping;
* `start == end` is **refused**. It reads equally well as "always open" and "never open", and an operator should not discover by experiment which one was implemented;
* the timezone is an IANA name validated **at save time** — a typo'd zone that failed later would hold every deploy for a window that never opens, surfacing days after the change;
* **DST in both directions**: on Rome's spring-forward day a `02:00`-`04:00` window opens at the local 03:00, because 02:00-02:59 does not happen; on the autumn day it is open across both passes of 02:00-02:59.

### The queue

`data/pending_deploys.json`, on disk because the window is hours away and a restart in between is ordinary. Written under a lock — events dispatch on daemon threads, so two certificates renewing at once write it concurrently.

| Situation | What happens |
| --- | --- |
| A second renewal before the window opens | One deploy, not two. The hook reads the certificate from disk when it runs. |
| The hook was deleted or disabled while waiting | Dropped. The current configuration says not to run it. |
| The window was removed | Released at the next drain, not one window later. |
| The hook fails inside its window | Recorded, **not** re-queued — otherwise it retries every minute for as long as the window is open. |
| Still waiting after 7 days | Logged as stuck. Not a limit — waiting is the feature — but a week means the days or the timezone are not what was meant. |
| The queue file is corrupt | Logged at error and the drain continues. A lost queue must not be silent. |
| An unreadable window (hand-edited file) | Deploys now. A disruption beats a certificate that never reaches the far end. |

Drained **every minute**, deliberately: a window is at least a minute long, so minute polling means no window can be missed. It is a true no-op when the queue file does not exist.

**Deploy Now** and `POST /api/deploy/test/<id>` ignore windows. Pressing the button is choosing this moment.

Held deploys are invisible everywhere else — the certificate renewed, the history shows nothing, nothing failed — so `GET /api/deploy/pending` lists them with their window and its next opening.

### Two defects found on the way

**Saving deploy hooks deleted every typed deploy target.** Settings → Deploy has no editor for `targets`: `loadConfig` never reads them, so `saveConfig` posts a body without them, and `save_config` assigned the whole `deploy_hooks` block. Opening that screen and pressing Save — *without touching anything* — silently removed every configured Kubernetes secret target. No error, no audit entry naming the loss, and the next renewal simply stopped publishing to the cluster. Absent now leaves alone; an explicit value, including an empty list, replaces. Same rule `CertificateService.update_config` already applies.

**The window fields were behind `x-show`**, which keeps the inner bindings live, so `hook.window.start` was evaluated for every hook *without* a window — one Alpine error per binding per hook in the console. `x-if` now.

### Checks

53 tests on the calendar, 29 on the queue, 9 on target preservation, 4 in a real browser. Each load-bearing decision was mutated to confirm a test fails: making the end inclusive, attributing a wrapped window to the day it ends on, dropping the equal-ends refusal, letting an unknown timezone become UTC in silence, ignoring the timezone entirely, and restoring the whole-block save. Full unit suite 3866 passed / 31 skipped, gated flake8 clean, coverage floors met, bandit and the theme codemod clean.

### DIFF
--- a/docs/deploy-hooks.md
+++ b/docs/deploy-hooks.md
@@ -9,10 +9,11 @@ This guide walks through:
 1. [What a hook is](#what-a-hook-is)
 2. [Configuring hooks (UI + JSON)](#configuring-hooks)
 3. [Environment variables passed to your command](#environment-variables-passed-to-your-command)
-4. [Manual triggering](#manual-triggering)
-5. [Security model: why some commands are rejected](#security-model)
-6. [Common recipes](#common-recipes)
-7. [Audit, history, and debugging](#audit-history-and-debugging)
+4. [Maintenance windows](#maintenance-windows)
+5. [Manual triggering](#manual-triggering)
+6. [Security model: why some commands are rejected](#security-model)
+7. [Common recipes](#common-recipes)
+8. [Audit, history, and debugging](#audit-history-and-debugging)
 
 ---
 
@@ -28,6 +29,7 @@ A hook is a JSON object with five fields:
 | `enabled` | boolean | no | Defaults to `true`. Disabled hooks are skipped during automatic firing but can still be tested manually. |
 | `timeout` | integer | no | Seconds. Default 30, capped at the system `MAX_TIMEOUT` (currently 300). |
 | `on_events` | string array | no | Subset of `["created", "renewed", "revoked"]`. If absent, the hook runs on all three. |
+| `window` | object | no | A [maintenance window](#maintenance-windows). If absent, the hook runs as soon as the certificate is issued or renewed — the behaviour every hook has had until now. |
 
 Hooks live under two keys in `deploy_hooks`:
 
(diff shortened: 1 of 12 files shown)


fabriziosalmi:
The `ui` job went red on this PR, in `test_ui_sso_failure_state.py` — which this PR does not touch. Re-running it failed identically, so it was not a flake. The cause is worth writing down, because it makes adding *any* UI test a landmine.

`browser_page` is **module-scoped** and logs in from scratch each time. The app rate-limits logins at **5 per IP per 60 seconds** (`modules/web/routes`), and the whole UI suite runs in about two minutes. There were five UI test modules. `test_ui_deploy_window.py` made six.

The sixth login inside that window was refused, so the fixture got no session cookie, every page redirected to `/login`, and the **last** module in alphabetical order failed with "the panel is empty" — while the change that broke it was an unrelated test file three modules earlier. The failing URL in the log says it plainly:

```
url='http://localhost:18888/login?next=/settings#oidc'
```

The login is now **session-scoped**: one login for the suite, whatever the number of modules. And a login that fails now fails the run with the status code and body, instead of yielding a cookie-less page — a browser suite silently exercising the login screen reports success while testing nothing.

Measured locally, full UI suite:

| | before | after |
| --- | --- | --- |
| result | 15 failed, 35 passed | **50 passed** |
| time | 8m40s | 1m12s |

The time is not an optimisation — the failures were 30-second Playwright timeouts.

Issue #608, opened 2026-08-28
A certificate with no private key is reported as healthy, and the scheduler skips it


fabriziosalmi:
## What happens

`get_certificate_info()` decides whether a certificate exists by looking at `cert.pem` alone. `privkey.pem` is never consulted:

https://github.com/fabriziosalmi/certmate/blob/main/modules/core/certificates.py#L1055-L1058

So a domain directory holding a valid `cert.pem` and **no private key** is reported as a healthy certificate:

```
domain           example.com
exists           True
expiry_date      2026-11-11 12:28:32
days_left        74
needs_renewal    False
```

That instance cannot serve TLS for `example.com`, and because `needs_renewal` is false the renewal scheduler will not touch it for another 44 days.

## Why this is reachable, not theoretical

This is the exact on-disk state produced by restoring a **masked backup** — the default — onto a fresh host. Since v2.26.0 a share-safe archive excludes key material (`_is_key_material`), so the restore writes back `cert.pem`, `chain.pem` and `fullchain.pem` without `privkey.pem`.

Reproduction, no mocks:

1. issue (or plant) a certificate, then `create_unified_backup(settings, include_secrets=False)`
2. restore that archive into an empty instance with `restore_unified_backup()`
3. `CertificateManager.get_certificate_info('example.com')` → the output above, with `certificates/example.com/privkey.pem` absent

A disaster-recovery operator following the obvious verification — "the API lists my certificates with sane expiries" — concludes the node is recovered when none of its certificates works.

## Suggested direction

Two separable pieces:

- **Report the truth.** `get_certificate_info` should surface the key state — e.g. a `private_key_present` (or `usable`) field — rather than letting `exists` imply a complete certificate. The dashboard and `/api/certificates` can then show a keyless certificate as broken instead of healthy.
- **Do not let it sit.** A certificate whose key is missing (or does not match `cert.pem`) should be treated as needing reissue, not as `needs_renewal: false`, so the scheduler acts instead of waiting for an expiry that is irrelevant.

A cheap correctness check for the match, which works for RSA and EC alike:

```bash
openssl x509 -noout -pubkey -in cert.pem | openssl sha256
openssl pkey -pubout -in privkey.pem     | openssl sha256
```

## Related

Found while verifying the disaster-recovery runbook against v2.26.1 (#607), which now documents the workaround: the runbook's §6 leads with the cert/key match check, because the API alone cannot tell an operator whether recovery succeeded.

PR #749, opened 2026-09-07
feat: issue and renew from a CSR, so the private key can stay on the device (#599)

### LINKED ISSUE #599: [Feature] CSR-only certificate issuance and renewal
Hi!

It would be useful if CertMate supported a CSR-only certificate workflow, where CertMate does not generate, store, or require access to the private key.

There are many appliances and security devices where the private key is generated and retained on the device and cannot be exported. These devices typically expose the CSR, but never the private key.

The desired workflow would be:

Generate the private key and CSR on the target device.
Submit the CSR to CertMate, either through the UI or via the CertMate API.
CertMate obtains the certificate from the configured CA/ACME provider using the supplied CSR.
CertMate manages the resulting certificate and certificate chain, without ever having access to the private key.
When the certificate needs to be renewed, CertMate can reuse the existing CSR and obtain a new certificate.
If the device performs key rotation, a new CSR can be submitted through the UI or API and associated with the certificate

Thx
Csszep

fabriziosalmi:
Closes #599.

> There are many appliances and security devices where the private key is generated and retained on the device and cannot be exported. These devices typically expose the CSR, but never the private key.

Pass the PEM as `csr` on the ordinary create endpoint, or paste it in the create form. CertMate obtains the certificate and manages the certificate and chain, and **never sees or stores a private key**.

```bash
curl -X POST https://certmate.local/api/certificates/create \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d "$(jq -n --arg csr "$(cat device.csr)" \
        '{domain: "api.example.com", dns_provider: "cloudflare", csr: $csr}')"
```

### Two facts about certbot's `--csr` mode — measured, not read

Verified end to end against Let's Encrypt staging with a real CSR before writing any of this:

* it writes `cert.pem`, `chain.pem` and `fullchain.pem` and **no private key**;
* it creates **no lineage** — no `renewal/` config, no `live/` tree. certbot says so itself on success:

  > Certificates created using --csr will not be renewed automatically by Certbot. You will need to renew the certificate before it expires, by running the same Certbot command again.

The second shapes everything else. `certbot renew` will never touch these, so renewal here **re-runs issuance with the stored CSR** — and "did it actually renew" cannot be answered by looking at `live/`. It is answered by fingerprinting the certificate before and after, which also means a CA returning the *same* certificate inside its reuse window is reported as "the certificate did not change" rather than advancing `renewed_at` past an expiry that stood still.

### The collision with #608

#608 made a certificate with no private key report `needs_renewal` — that is what restoring a share-safe backup produces, and it used to report itself healthy. A CSR-only certificate is **deliberately** keyless. Without a fourth state it would be reissued on **every renewal sweep**, burning CA rate limit for a file that must not exist.

`private_key_state` gains `external`:

| State | `private_key_present` | `usable` | forces `needs_renewal` |
| --- | --- | --- | --- |
| `present` | true | true | no |
| `missing` | false | false | **yes** (#608) |
| `mismatched` | true | false | **yes** |
| `unknown` | null | null | no |
| **`external`** | **false** | **null** | **no** |

`usable` is null rather than false: the certificate *is* usable, on the device that holds the key. CertMate cannot answer for a key it does not have. `_private_key_present` and `_usable` are now written out per state rather than derived from a comparison, because five states answer those two questions three different ways.

### Refused rather than silently reinterpreted

The names come from the CSR — it is signed over its own subject and SANs, so the CA issues for those whatever the request said.

| Situation | Result |
| --- | --- |
| `csr` + `san_domains` | 400. The certificate covers the CSR's names; a request naming others would get something different from what it asked for. |
| `csr` + `key_type`/`key_size`/`elliptic_curve` | 400. The device generated the key. |
| The CSR does not cover `domain` | Refused — the directory is named for it and renewal, the health check and every deploy hook find it by that name. |
| The signature does not verify | Refused before anything is written. The CA would reject it after CertMate had created a directory and reported progress. |
| `domain` already has a managed private key | Refused. Converting would leave the old `privkey.pem` beside a certificate it cannot serve. |
| A CSR name outside the key's `allowed_domains` | Refused per name, the way SANs already are. |

The create form does not send the conflicting fields at all — it disables the SAN editor and the key selector once a CSR is present — so that refusal is not something an operator discovers after an ACME round trip.

### Deploying one

* **Shell hooks** run normally. `CERTMATE_KEY_PATH` is **not set** rather than pointed at a file that does not exist — the path would be a statement that is not true, and a hook can ask `[ -n "$CERTMATE_KEY_PATH" ]` and get a truthful answer. Nothing changes for a key-managed certificate, which is every existing one.
* **Typed deploy targets** all publish the key with the certificate, so none can serve one. They now say exactly that, naming the reason and what to use instead, rather than the generic "certificate files unreadable" they would otherwise report on every renewal.

### Checks

44 unit tests, 7 in a real browser, plus the real-CA verification above. Thirteen mutations, **all killed**: dropping the signature check; keeping each flag that must go (`-d`, `--cert-name`, `--key-type`); dropping ones that must stay; losing the name dedup; letting an external key read as `missing`; making `usable` false; accepting a CSR for other names; allowing the conversion; routing the renewal through `certbot renew`; reporting an unchanged certificate as renewed.

Full unit suite 3910 passed / 31 skipped, gated flake8 clean, coverage floors met, docs-navigation gate green.

**One thing left open on purpose:** `docs/csr-only-certificates.md` is English only. The four translated trees are already three pages behind (16 vs 19), which is the project's existing state rather than something this PR introduces — flagging it rather than shipping four unreviewed translations of a technical document.


fabriziosalmi:
The real-CA E2E is green, and getting it there found **five** defects that all 46 unit tests agreed were fine. Recording them, because the pattern matters more than any one of them.

| # | Defect | Why no unit test saw it |
| --- | --- | --- |
| 1 | `private_key_state` was `unknown`, not `external` | Every unit test drove the filesystem branch of `get_certificate_info`. A real instance goes through the **storage** branch — the local backend is a storage backend too. |
| 2 | Every CSR renewal was a **500** | `create_certificate` refuses a domain that already has a certificate, and a renewal by definition does. The unit tests all replaced `create_certificate` with a double, so the guard they were driving past did not exist in any of them. |
| 3 | **certbot refuses to overwrite its own output** in `--csr` mode | `FileExistsError: … csr-out/cert.pem`, *before* contacting the CA. The refusal is certbot's, so a stubbed executor can never produce it. The **first renewal of every CSR certificate** failed: issuance perfect, then silence until expiry. |
| 4 | The renewal returned `(bool, str)` where the route reads `result.get('renewed')` | The shape was **invented rather than copied** from the ordinary renewal — and every unit test then asserted that invention. They agreed with each other and with nothing else. |
| 5 | A run finished `1 skipped` with nothing saying why | My own skip helper. A silent skip is indistinguishable from a pass. |

Each now has a test that bites, and two of them are tests of a *contract* rather than of a branch: one reads `result.get('renewed'` off the route's own source, the other asserts the precondition certbot cares about (its output files do not exist when it starts) since the refusal itself is unreachable from a double.

### What the E2E proves

Against Let's Encrypt **staging**, with a CSR generated outside CertMate:

* a real certificate is issued, covering **exactly** the CSR's names;
* `GET …/download/privkey` returns 404 — there is no key on the node;
* the certificate reports `private_key_state: external`, `usable: null`, `needs_renewal: false`;
* a forced renewal re-issues **from the stored CSR** and still leaves no key.

Cert burn: 2 staging certificates per run.

Full unit suite 3918 passed / 31 skipped, gated flake8 clean, coverage floors met.

PR #789, opened 2026-09-08
fix: the renewal sweep names the certificates it cannot renew (#759)

### LINKED ISSUE #759: certificate auto renew Private CA
**Describe the bug**
When a certificate is generated from a private CA with a lifespan of 31 days and a threshold of 30, the renewal will not execute automatically (it was executed manually on the 25th day). I am not sure if this is because the threshold was so close to the lifespan, or if it is because the private CA has no auto-renewal configured. 

Could you please check that? 

PS. There was not any evidence of auto renew and fail in log as well, log didn't have anything for that domain except creation and deploy hooks that I run when it was created.

Thank you.

**To Reproduce**
1. Generate certificate with lifespan 31 days (or 32)
2. wait
3. auto renew is not executed (manual works)

**Expected behavior**
auto renew triggered 

linux kubernetes/docker

fabriziosalmi:
Closes #759.

## The threshold is not the cause

The report suspects the 31-day lifespan against the 30-day threshold, or the
private CA. Reproduced against the real code at eight points across the
certificate's life, a **registered** certificate with exactly those numbers is
picked up from day zero — and `ca_provider` never enters the decision:

```
 giorno  days_left  needs_renewal  esito dello sweep
      0         30           True  checked=1 tentato_rinnovo=1
      1         29           True  checked=1 tentato_rinnovo=1
     25          5           True  checked=1 tentato_rinnovo=1
```

`days_left` is 30 on the day it is issued and the comparison is `<=`, so it
renews immediately.

## What actually happened

The detail that matters is in the report, not the title: **the log carried
nothing at all for that domain.** Not a failure, not a skip, not a "not yet
due". Reproducing that:

```
CERTIFICATO SUL DISCO MA NON IN settings['domains']
  il cert esiste:        True, days_left=5
  needs_renewal:         True   <- dice SI
  lo sweep lo visita:    checked=0
  rinnovi tentati:       0
  righe di log sul dominio: NESSUNA
  righe di log dello sweep: ['Checking 0 certificate(s) for renewal']
```

Exact match, including "manual renewal works" — the manual path is addressed by
domain and never consults the list.

## The two views disagree

`check_renewals` iterates `settings['domains']` and nothing else. The rest of
the application does not:

| Component | Which certificates it considers |
|---|---|
| `digest.py` | **union** of `settings['domains']` and the certificate directories |
| `get_certificate_info` | reads the disk |
| `check_renewals` | `settings['domains']` only |

So the certificate shows on the dashboard with `needs_renewal: true`, the digest
reports it expiring — and the one component whose job is to renew it never
mentions it. That silence is why the report reads as a threshold bug: there was
no other evidence to read.

## The change

The sweep reconciles what it considered against the certificate directories.
Each one no settings entry named gets a warning naming it and the remedy, plus a
new `unmanaged` count carried in the completion line:

```
Certificate 'orphan.example.com' exists on disk but no settings entry names it,
so this renewal sweep did not consider it and it will not renew. Re-register it
with the 'Add Domain' flow or POST /api/settings.
```

**It reports; it does not adopt.** Renewing a certificate the operator never
registered would issue against a CA for a domain CertMate was not asked to
manage — a half-restored backup would start requesting certificates — and a
background sweep is not where that decision belongs. Happy to add self-healing
adoption as a follow-up if you want it; it is a policy call, not a bug fix.

## Deliberately not done

- **The `settings.py` warning stays as it is.** It only fires when `domains` is
  entirely empty, which is exactly why an instance with nine registered
  certificates and one orphan was told nothing. Broadening it would put a
  `scandir` on every `load_settings()` call; the sweep now emits a better
  message once per run instead.
- **No new Prometheus gauge.** `record_renewal_sweep` is part of the `/metrics`
  surface and widening it belongs in its own change.

## Found while writing the tests

`load_settings` **strips** a domains entry it cannot read a domain out of — both
a dict with no `domain` key and a bare number come back as `domains: []`. That
makes the sweep's own `skipped_invalid` branch unreachable through normal
loading, and it means a typo in a hand-edited `settings.json` used to remove a
certificate from renewal with nothing said anywhere. The reconciliation catches
that case too, and a test records it.

## Verification

10 new tests; **8 fail against the code they replace**, and the 2 that pass are
the diagnosis (they describe behaviour that was already there). Controls
included: a registered certificate is not reported, an `auto_renew: false`
certificate is not reported (a warning that fires on a correct configuration is
one people learn to skip), and an unreadable certificate directory does not cost
the sweep its work.

Locally 4434 passed, 0 failed across the suite excluding the seven
container-building files.

### DIFF
--- a/modules/core/certificates.py
+++ b/modules/core/certificates.py
@@ -30,7 +30,9 @@ from pathlib import Path
 from cryptography import x509
 from .shell import ShellExecutor
 from .dns_strategies import DNSStrategyFactory, HTTP01Strategy, acme_webroot_dir, check_certbot_plugin_installed
-from .constants import METADATA_SCHEMA_VERSION, CERTIFICATE_FILES, DEFAULT_RENEWAL_THRESHOLD_DAYS
+from .constants import (METADATA_SCHEMA_VERSION, CERTIFICATE_FILES,
+                        DEFAULT_RENEWAL_THRESHOLD_DAYS,
+                        iter_cert_domain_dirs)
 from .structured_logging import LogContext, new_correlation_id
 from .csr_issuance import (
     CSR_OUTPUT_DIRNAME, CSR_OUTPUT_FILES, CSRError, csr_domains,
@@ -3304,9 +3306,15 @@ class CertificateManager:
 
         if not settings.get('auto_renew', True):
             logger.info("Automatic renewal is globally disabled; skipping renewal check")
+            # `unmanaged` present but zero: the key is part of the summary
+            # shape, and a caller that reads it must not have to know which
+            # early return produced the dict. Nothing is reported here because
+            # renewal was switched off deliberately — naming certificates the
+            # sweep "did not consider" when it considered none would be noise.
             return {'checked': 0, 'renewed': 0, 'failed': 0,
                     'skipped_disabled': 0, 'skipped_invalid': 0,
-                    'skipped_not_due': 0, 'auto_renew_disabled': True}
+                    'skipped_not_due': 0, 'unmanaged': 0,
+                    'auto_renew_disabled': True}
 
         # Migrate settings format if needed
         settings = self.settings_manager.migrate_domains_format(settings)
@@ -3316,7 +3324,12 @@ class CertificateManager:
 
         summary = {'checked': 0, 'renewed': 0, 'failed': 0,
                    'skipped_disabled': 0, 'skipped_invalid': 0,
-                   'skipped_not_due': 0}
+                   'skipped_not_due': 0, 'unmanaged': 0}
+        # Every domain this sweep took a decision about, so the reconciliation
+        # below can name the certificates it never reached. Collected rather
+        # than re-derived from `domains`, because a malformed entry is skipped
+        # here and must not read as "seen".
+        considered = set()
         started = time.time()
         self._report_unfinished_sweep()
         self._mark_sweep_started(len(domains))
@@ -3343,6 +3356,8 @@ class CertificateManager:
                     logger.warning(f"Skipping domain entry with no domain name: {domain_entry!r}")
                     summary['skipped_invalid'] += 1
                     continue
+
+                considered.add(domain)
 
                 # Per-certificate opt-out: skip when auto_renew is explicitly
                 # disabled on this domain entry. The global auto_renew flag is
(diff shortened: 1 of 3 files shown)


fabriziosalmi:
## End-to-end against a real private CA

Run on a throwaway Proxmox LXC: **step-ca** as the private ACME CA (leaf lifetime
20 minutes), CertMate built from this branch, `http-01` served through nginx on
port 80, everything on an internal Docker network — no external DNS, no
credentials, nothing published.

### 1. A private-CA certificate issues, and it renews on its own

```
create -> 201 {"domain": "e2e.internal", "ca_provider": "private_ca", "duration": 3.43}

issuer=O=E2E, CN=E2E Intermediate CA
serial=057C86E5567229EC338821AAD7A9BFCB
notBefore=Sep  8 16:23:23 2026 GMT
notAfter =Sep  8 16:44:23 2026 GMT        <- 21 minuti
days_left=0 needs_renewal=True ca=private_ca
```

The **real APScheduler job** then fired on its own (the bench runs it every two
minutes instead of 02:00; nothing else was changed):

```
16:26:00  Renewing certificate for e2e.internal
16:26:03  Successfully renewed certificate for e2e.internal
16:26:03  Renewal check complete in 3.4s: 1 checked, 1 renewed, 0 failed,
                                          0 disabled, 0 invalid, 0 not-due, 0 unmanaged

serial=CC30ED7050C9A77C437F19F2750FFFBF   <- cambiato
```

So **auto-renewal of a private-CA certificate is not broken.** A registered
certificate with a lifespan shorter than the threshold renews, against a real
ACME CA, through real certbot. That is the control the diagnosis needed.

### 2. The reported condition, reproduced with real components

Same instance, same certificate, one change — the domain removed from
`settings['domains']` while the certificate stays on disk:

```
16:28:00  WARNING  Certificate 'e2e.internal' exists on disk but no settings
                   entry names it, so this renewal sweep did not consider it and
                   it will not renew. Re-register it with the 'Add Domain' flow
                   or POST /api/settings.
16:28:00  Renewal check complete in 0.0s: 0 checked, 0 renewed, 0 failed,
                                          0 disabled, 0 invalid, 0 not-due, 1 unmanaged

serial=CC30ED7050C9A77C437F19F2750FFFBF   <- invariato
```

`0 checked` with the certificate sitting there expiring — that is exactly the
"no evidence in the log" the report describes, and it is now one warning naming
the domain and the remedy. The serial is unchanged, which is the other half of
the design: it reports, it does not adopt.

### Incidental, and worth keeping

Two attempts to bring the bench up were refused by the token validation before
anything served — first for length, then for containing `12345`, each time with
"Refusing to serve: ignoring it would leave this instance with NO
AUTHENTICATION". Working exactly as intended on a real deployment.

The `request_id` / `operation: renewal_sweep` fields in every line above come
from the correlation-id work; this is the first time they have been seen doing
their job on a live sweep.

